PR #31, opened 2016-02-25
Changed build.bat to copy Zing files to the bin dir and to use the pl…

paulthomson:
…atform var. Added some files to .gitignore.

(This ensures the copied files are correctly ignored by zing's .gitignore. )

### DIFF
--- a/Bld/build.bat
+++ b/Bld/build.bat
@@ -8,6 +8,9 @@ if "%Configuration%"=="" set Configuration=Debug
 set Platform=%2
 if "%Platform%"=="" set Platform=x86
 
+echo msbuild  P.sln /p:Platform=%Platform% /p:Configuration=%Configuration% /t:Clean
+msbuild  P.sln /p:Platform=%Platform% /p:Configuration=%Configuration% /t:Clean
+
 git submodule init
 git submodule update
 cd ext\zing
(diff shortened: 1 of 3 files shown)


shazqadeer:
Could you explain what problem this  modification is trying to solve?

paulthomson:
Yes. Sorry.

* In my git client, binaries in `Ext\Zing\$(Platform)` were showing up and could not be ignored via P's `.gitginore`. This could be partly due to my git client but by putting them in `Ext\Zing\bin\$(Platform)` then they will definitely be ignored because Zing's `.gitignore` includes `bin/` (i.e. `Ext\Zing\bin`).
* The `build.bat` file was only `x86` and `x64` versions of Zing (depending on the chosen platform) but then it only copied the `x86` version (sometimes into a folder called `x64`) which might not even exist. I changed it to copy the version that was built to the correctly named directory. 
* Some files were missing from `.gitignore`.

shazqadeer:
@clovett : Please review and accept if you are satisfied.

clovett:
NOTE: I had an idea on how to improve things, I moved the path to zc.exe and zinger.exe into P.props and modified the above projects to use that msbuild property instead:
        <ZCompiler>$(MSBuildThisFileDirectory)Bld\$(UnifiedPlatform)\zc.exe</ZCompiler>
        <Zinger>$(MSBuildThisFileDirectory)Bld\$(UnifiedPlatform)\zinger.exe</Zinger>

but the C# code in program.cs will still need to be fixed.

paulthomson:
Sounds good! Do you want to push that or should I do the above?

paulthomson:
@clovett OK I think I have done this.

paulthomson:
> Moving the Zing build outputs to the "Bld" folder makes sense, but perhaps we should then move the binaries into the full "Bld\Drops\Debug\x86\Binaries" path ? Note also in the master branch I already created "" and "" so we need to merge that, I also made some other build related fixes in master. Can you merge master and then do another pull request?

I have made some new commits. Note that I moved the "clean" up in `build.bat` to before copying zing files as otherwise it would remove the copied zing files. See what you think.

shazqadeer:
Are we done with this one?  If yes, please close.

paulthomson:
Cool. Looks like @clovett has taken care of this in his latest commit.

PR #114, opened 2017-04-19
Fix a couple of issues with PingPong sample.

tvrprasad:
- PingPong.exe crashes when run as "PingPong.exe" without a / or \ in the
path.
- PingPong.exe was getting copied to PingPongFast instead of PingPong
directory.


shazqadeer:
Done.  Thanks.

PR #131, opened 2017-05-25
push $SCRIPTPATH before building runtime

ficoos:
This makes the build script executable from outside the Bld dir. It was
the only part of the build that didn't have this property and there is
no real reason not to have it.


ankushdesai:
Thanks for fixing this.

PR #132, opened 2017-05-25
Allow using msbuild instead of xbuild

ficoos:
xbuild has been deprecated and msbuild is available since Mono 5.0

- This patch makes the build script check for msbuild before falling
  back to xbuild.
- It also allowes forcing the build program in the environment.


ankushdesai:
thanks !!

PR #137, opened 2017-06-19
Add NUnit test harness to project

alexreinking:
This PR contains an NUnit unit test that is compatible with ReSharper or the built-in testing GUI with the (free) NUnit 3 Test Harness extension installed. This makes it easy to selectively re-run tests, and will enable easy parallelization down the line. We also don't have to drop to the command line. No changes need to be made to the tests.

Some code is duplicated with RunPTool, which is used by the batch file tester. This is to make it easier to remove RunPTool as we deprecate the old tester. The main remaining "TODO" is to add an option to freeze the test output.


ankushdesai:
I am getting the following error when I run NUnit test using ReSharper.
Can you please check:
image

alexreinking:
Unable to repro from a clean repository. Do you have any special environment configuration?

alexreinking:
Be sure you have disabled shadow copying in ReSharper's settings and have msbuild.exe in your path (either add it manually in system settings or open VS from the developer prompt by "running" `P.sln`)

ankushdesai:
Thanks, those two things fixed the problem. 
I am able to run the regressions using NUnit + ReSharper.

PR #149, opened 2017-08-21
Removed support for extern declarations

ankushdesai:
extern on machine and function declaration is no longer supported


ankushdesai:
@shazqadeer Extern has been eliminated.

PR #164, opened 2018-02-20
Modp

ankushdesai:
Lets merge the module system !


Src/Pc/CompilerCore/TypeChecker/ModuleSystemTypeChecker.cs
```diff
@@ -0,0 +1,25 @@
...
+        private readonly ITranslationErrorHandler handler;
+
+        private ModuleSystemTypeChecker(ITranslationErrorHandler handler)
+        {
+            this.handler = handler;
+        }
+
+        public static void CheckWellFormedness(ITranslationErrorHandler handler, IPModuleExpr moduleExpr)
+        {
+            moduleExpr.CheckAndPopulateAttributes(handler);
```
alexreinking: All the `CheckAndPopulateAttributes` functions checked _before_ constructing a `ModuleExpr`, but could also be moved here for validation. In general, we want to validate things before we commit them to immutable objects, and in the worst case we can check validity just before returning them. But all validity checks should be external, since fusing those checks to the objects makes them harder to change.

ankushdesai: The problem I was facing was that these functions access some of the private information of these objects for computing/checking. So should I make the function local or the information public?
I chose the former.
While I was implementing this class, it did occur to me that I was deviating from the way other code is structured.

alexreinking: I expect that we'll need to access those fields in the code generator down the line, so they should at least be get-only properties (as opposed to public fields) that are set in the constructor. 

The underlying issue is that `ModuleExpr` <: `IPModuleExpr` <: `IPAST` which implies a chain of "is-a" relationships between the classes. Compare this to the implementations of EventSets and PLanguageTypes in the library. AST nodes _have_ event sets and they _have_ types; they aren't AST nodes themselves. ModuleExprs should probably _have_ objects that represent the constraints derived from the syntax. Additionally, it is strange for data objects (like AST nodes) to throw exceptions, which are meant to communicate logical failures (ie. issues with the construction of an object or violated function parameter invariants).

A good indicator that code needs refactoring is the presence of a **constant field**. The `IPModuleExpr` member `bool IsWellFormed { get; }` is always true post-analysis, since otherwise an exception is thrown. It is only used to control the recursive calls to `CheckAndPopulateAttributes` which really just emulates a `Lazy<bool> IsWellFormed { get; }` whose value is set according to the logic in "Check...". 

But since that `Lazy` would still be constant (ie. if it's false it either hasn't run yet or you throw an exception) there's no point in having it be part of the object. It would be best to either delay creating the object until you know that it will be valid, or externally validate it before returning it.

Src/Pc/CompilerCore/TypeChecker/AST/IPModuleExpr.cs
```diff
@@ -0,0 +1,58 @@
...
+        IEventSet PrivateEvents { get; }
+        IInterfaceSet PrivateInterfaces { get; }
+        IEventSet Sends { get; }
+        IEventSet Receives { get; }
+        IInterfaceSet Creates { get; }
+
+        //used for code generation and runtime
+        IDictionary<Interface, IDictionary<Interface, Interface>> LinkMap { get; }
+        IDictionary<Interface, Machine> InterfaceDef { get; }
+        IDictionary<Machine, IEnumerable<Interface>> MonitorMap{ get; }
```
alexreinking: maybe use `IReadOnlyList` instead of `IEnumerable`?

PR #165, opened 2018-03-21
Cruntime

ankushdesai:
Done with the PRT_SM_FUN changes.
Note that for receive:
I have added a field: packedReceiveEvents what is initialized to the receive events if blocked on receive and to NULL otherwise. The code for receive-blocking and unblocking is commented out right now. We need to add Dan libhandler logic there.


alexreinking:
Looking good. Thanks for turning this around!

PR #213, opened 2019-08-23
Add 'break' and 'continue' statements

exists-forall:
This adds basic support for `break` and `continue` statements to P, and translates these statements directly to their equivalents in the target language.

One shortcoming of this implementation is that the P compiler does not currently attempt to detect when a `break` or `continue` statement is used outside a loop.  If the user attempts to use such a statement outside a loop, the P compiler will run without errors, but it will generate code in the target language which does not compile.  Ideally, the P compiler should never generate malformed code, and should instead include a simple analysis pass to catch these kinds of errors earlier.


exists-forall:
The issue described above has been fixed; the compiler now generates an error message when the user attempts to use a `break` or `continue` statement outside a loop body.

ankushdesai:
This is great work @exists-forall, two final things before we merge the request.
1) Create a couple of test-cases (regressions) using break and continue.
2) Make sure that both Prt and PSharp regressions pass.

exists-forall:
I have added regression tests for compile-time error handling and for the runtime behavior of `break` and `continue` on the P# target.  I am still having trouble running the `Prt` tests on my machine, so I have not included any runtime tests for the `C` target.

During testing, I discovered a bug in the code generation for `continue` on both targets, which I have fixed by changing how `while` statements are simplified in the `IRTransformer`.

PR #218, opened 2019-09-07
In operator for list for both C and C# targets

paulshaoyuqiao:


ankushdesai:
All this looks great. @paulshaoyuqiao Can you please add regressions for the "in" operator.

PR #223, opened 2019-09-21
Add short-circuiting behavior for boolean operators `&&` and `||`

exists-forall:
This change modifies the IR transformer to transform`&&` and `||` into code which uses `if` statements to implement short-circuiting behavior.  It turns out the test suite already contained a regression test to check the short-circuiting behavior of boolean operators, but the code to test it was intentionally commented out until the feature was implemented.  Rather than create a new regression test, I uncommented the code in this existing test.


ankushdesai:
This is great, thanks for the quick fix @exists-forall!!

PR #226, opened 2019-10-13
Regression fix

Eric-Le-Ge:
Fixes regression tests for both Dynamic and Correct test cases so that they do not require a golden output file.

TBD: Compiler (static) tests currently don't support exit code. For those tests, would we like to extract and compare exit codes or compare the entire stdout and stderr with golden output?


Tst/UnitTests/RegressionTests.cs
```diff
@@ -81,7 +81,7 @@ public void TestTemp()
     public class PrtSharpRegressionTests
     {
         private static IEnumerable<TestCaseData> RegressionTestSuite =>
-            TestCaseLoader.FindTestCasesInDirectory(Constants.TestDirectory, new[] {"Prt", "PrtSharp" });
+            TestCaseLoader.FindTestCasesInDirectory(Constants.TestDirectory, new[] {"DynamicError", "DynamicErrorPrtSharp", "Correct" });
```
ankushdesai: Is DynamicErrorPrtSharp a separate folder?

Eric-Le-Ge: Yes, there are a few cases that seem to be written for only C# or C runtime. I modified the parsing logic to and renamed few folders to address this need. 

Also, static errors are platform independent, so they are separate from either PrtRegressions or PrtSharpRegressions

ankushdesai: Can you please point out these cases? I would like to look at these and then fix them in some ways to make it uniform.

Eric-Le-Ge: Yes, there are only 2 such cases which are located in this folder:
Feature2Stmts/DynamicErrorPrtSharp/

ankushdesai: Thanks, Eric, for identifying these (even the unused) cases.

Can I request you to add DynamicError* testcases to DynamicError and Correct* to Correct.

Please let me know if they dont pass for Prt or PrtSharp.

Eric-Le-Ge: Yes, actually, all the *Unused cases are not passing which is why they are included in a separate folder. Current set of StaticError, DynamicError and Correct is the maximum passing set

ankushdesai: Thanks Eric, for all this experiments. Can I ask you to do the following final change:
Create the appropriate folders such that the *Unused cases also pass and are included in the regressions.
I think most of them will pass the *PrtSharp.

ankushdesai:
This is superb work, and thanks to you, the regressions are finally all cleaned up.
There are no stale (unused) regressions.

I am merging the changes, looking forward to the next pull-request for the set type.

PR #230, opened 2019-10-25
set implementation

Eric-Le-Ge:
Implements the Set data type in P
An example of its usage can be found in test program at Tst/RegressionTests/Feature4DataTypes/Correct/SetImpl/SetImpl.p


Src/Pc/CompilerCore/Parser/PParser.g4
```diff
@@ -143,6 +144,7 @@ statement : LBRACE statement* RBRACE                      # CompoundStmt
           | CONTINUE SEMI                                 # ContinueStmt
           | lvalue ASSIGN rvalue SEMI                     # AssignStmt
           | lvalue INSERT LPAREN expr COMMA rvalue RPAREN SEMI # InsertStmt
+		  | lvalue INSERT LPAREN rvalue RPAREN            # AddStmt
```
ankushdesai: @Eric-Le-Ge Is there a missing `SEMI` in the end?

PR #237, opened 2019-11-21
String type

Eric-Le-Ge:
Implementation of String Type.


Src/CoyoteRuntime/Values/PrtString.cs
```diff
@@ -0,0 +1,110 @@
...
+                return false;
+            }
+
+            return obj is PrtString other && Equals(other);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public IPrtValue Clone()
+        {
+            return this;
```
ankushdesai: Shouldnt this be new PrtString(this.val)?

Src/PSharpRuntime/Values/PrtString.cs
```diff
@@ -0,0 +1,110 @@
+﻿using System;
```
ankushdesai: This file should not be around? The folder PSharpRuntime is deleted.

Src/CoyoteRuntime/Values/PrtString.cs
Eric-Le-Ge: since P strings are not mutable sharing a pointer should be fine

Src/PSharpRuntime/Values/PrtString.cs
Eric-Le-Ge: Yes, I'll delete this

Src/CoyoteRuntime/Values/PrtString.cs
ankushdesai: Agreed!

PR #240, opened 2019-12-04
upgrade coyote

pdeligia:
@ankushdesai this PR updates P to Coyote `rc7` which introduces typed transitions for state machines: `RaiseEvent`, `GotoState`, `PushState` and `PopState` all now return a `Transition` (accordingly for a monitor). I am using the same approach that you do for monitors to handle these at the top level handlers and return back the transition. This is most likely the last major API change in Coyote as we are trying to polish and stabilize the APIs.


ankushdesai:
The build is failing on a single test case. Let us also discuss these changes.

pdeligia:
> The build is failing on a single test case. Let us also discuss these changes.

Everything passes now!

Src/Pc/CompilerCore/Backend/Coyote/CoyoteCodeGenerator.cs
```diff
@@ -775,6 +757,11 @@ private void WriteStmt(CompilationContext context, StringWriter output, IPStmt s
                     break;
 
                 case FunCallStmt funCallStmt:
+                    if (funCallStmt.Function.CanChangeState == true || funCallStmt.Function.CanRaiseEvent == true)
```
ankushdesai: @pdeligia I just realized this thing would not work out. Consider the function:
```
foo() {
if($) 
  goto S;
}
```
In other words, we will always early terminate on a function call that can change state or raise an event in its function body.

pdeligia: Good catch! This should be now fixed.

ankushdesai:
Thanks for these changes.

PR #244, opened 2020-01-28
Update PLogFormatter to new IActorRuntimeLog.

lovettchris:
**Don't merge this until the next Coyote update...** I'm just giving folks a heads up... `IActorRuntimeLogFormatter` is going away.  You can still implement your own entirely custom `IActorRuntimeLog` like we did with the `ActorRuntimeLogGraphBuilder`, or you can subclass  a new "text logging" implementation of this interface called `ActorRuntimeLogTextFormatter`.  The next coyote update also provides a new "--xml-trace" option which uses a new `IActorRuntimeLog` implementation called `ActorRuntimeLogXmlFormatter`.  This replaces the old crufty *.pstrace files with a much more machine readable xml format which might also be handy for any tooling you build around these logs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -233,7 +233,7 @@ stubs.c
 
 Src/PrtDist/Binaries/clusterconfiguration.xml
 *.idb
-/Ext/PsTools
+/Ext
 /Src/Prt/Samples/ForeignTypes/Tester.ipdb
 /Src/Prt/Samples/ForeignTypes/Tester.iobj
 /Src/Prt/Samples/ForeignTypes/Tester.exe
--- a/Src/CoyoteRuntime/CoyoteRuntime.csproj
+++ b/Src/CoyoteRuntime/CoyoteRuntime.csproj
@@ -5,7 +5,7 @@
     <OutputPath>D:\Workspace\P\Bld\Drops\Debug\Binaries</OutputPath>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Coyote" Version="1.0.0-rc7" />
+    <PackageReference Include="Microsoft.Coyote" Version="1.0.0-rc9" />
     <PackageReference Include="System.Collections.Immutable" Version="1.5.0" />
   </ItemGroup>
 </Project>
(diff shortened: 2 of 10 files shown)


ankushdesai:
Thanks @lovettchris!
I had a request, @pdeligia can we also include the suggested change of adding `Event` as a payload to the log functions instead of just `eventName`. Then one can log whatever information information they need for the event (for P the attached payload).

lovettchris:
@ankushdesai , Event e is done.  In the process I removed the code that was doing stateName.Split(".").Last() and eventName..Split(".").Last().  This kind of thing will lead to ambiguous logs is the "Tail" part of a name is the same across two events or states.  For example, you might have "Microsoft.Samples.CoffeeMachine.Init" state and the "Microsoft.Samples.SafetyMonitor.Init" and you don't want to confuse these two states just because they both have the same name "Init".  Same goes for events.

ankushdesai:
@lovettchris Thanks for your changes.
But can you revert your changes that removed the Split(".") operation.

> Assuming the names are local to the Machine Type is not always correct either, especially for Events which might be defined outside the class, e.g. global events like "Microsoft.Coyote.Actors.HaltEvent" will not have the same base name as your custom StateMachine type.

In the context of P language, states are always enclosed inside a machine and also all events are always global events. Hence, that change was valid and is important for generating readable P traces.

If we remove `Split(".")` then the log statements refer to events in generated code with `Microsoft.*` which does not really exist in the P language level and confuses the developers.

The build is failing right now as the regressions are failing. If you cannot fix these, would it be possible for you to merge your changes in a branch so that I can work on that branch and then merge it into the master.

Thanks again for this PR.

lovettchris:
@ankushdesai , done, short names are back.  Doing so for the new strongly typed "Event e" arguments required copy/paste actual `ActorRuntimeLogTextFormatter` implementations over to this class.  Note, I have not tested this works correctly, but the copy past was pretty mechanical so it should work. You could of course copy all the implementation from `ActorRuntimeLogTextFormatter`  and go to town with changing the actual text formatting to whatever you want...

lovettchris:
Ok, now I have tested that it works, on the TwoPhaseCommit example, here's the log.
[TwoPhaseCommit_0_0.txt](https://github.com/p-org/P/files/4132723/TwoPhaseCommit_0_0.txt)

ankushdesai:
Thanks @lovettchris , I will take it from here and fix the regressions that are failing.

PR #246, opened 2020-02-15
Coyote is hiding the Transition type. 

lovettchris:
There is no reason to expose Transition objects and force users to "return" a Transition.  As is evidenced by the dangling Transition PR, it didn't actually improve things and the solution to dangling Transitions shows that exposing Transition is not required to ensure the same contract (namely, that only one is created per action) as shown by the internal PendingTransition member variable.

We also decided to rename some methods to make them all have a similar and therefore consistent name:

```
RaiseEvent is unchanged
Rename “GotoState” to “RaiseGotoStateEvent”
Rename “PushState” to “RaisePushStateEvent”
Rename “PopState” to “RaisePopStateEvent”
Rename “Halt” to “RaiseHaltEvent”
```

The idea being that the "Raise...Event" language more clearly communicates to the user that these operations are queuing something up to happen after the action completes and they are not processed immediately.  This also makes it easier for the user to remember the idea that currently you can only do one Raise operation per action.

I tried to fix the CoyoteCodeGenerator, and most of the PrtSharpRegressionTests are passing, but occassionally I'm seeing an error about access denied trying to copy CoyoteRuntime.dll saying the file is in use.  I added some retry logic to the File.Copy operation in CoyoteRunner.dll but it still seems to happen.  Perhaps we just need to increase the number of retries, or investigate more deeply why this is happening, but I don't think it is the result of this PR.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,14 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
-  <activePackageSource>
-    <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
-  </activePackageSource>
   <packageSources>
+    <clear/>
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
     <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
+    <add key="Coyote" value="../../foundry99/Coyote/bin"/>
   </packageSources>
   <config>
     <!-- Defend against repositoryPath being set somewhere else in the environment. -->
-    <add key="repositoryPath" value="packages" />
+    <add key="globalPackagesFolder" value="./packages" />
+    <add key="repositoryPath" value="./packages" />    
   </config>
 </configuration>
--- a/NuGet.config
+++ b/NuGet.config
@@ -4,7 +4,6 @@
     <clear/>
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
     <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
-    <add key="Coyote" value="../../foundry99/Coyote/bin"/>
   </packageSources>
   <config>
     <!-- Defend against repositoryPath being set somewhere else in the environment. -->
--- a/P.sln
+++ b/P.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2036
+# Visual Studio 16
+VisualStudioVersion = 16.0.27130.2036
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Runtime", "Runtime", "{D641DC13-F2F0-4365-963A-DCFC80BABCAA}"
 EndProject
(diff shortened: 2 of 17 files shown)


ankushdesai:
@lovettchris In P, unlike in P#, raise and goto statements can happen at any point in the handlers.
They lead to early termination of the function i.e., statements after raise and goto are not executed. @pdeligia had handled this scenario by adding a local variable that stores the return condition.
I am not sure if your current change handles this case correctly.

@pdeligia can you please confirm, if we are reverting back to the previous implementation. Should we also return back to the implementation with throwing an exception?

pdeligia:
@ankushdesai yep, we are reverting back to our previous implementation, as Chris found a nice way to keep `Transaction` internal, and still give an error if someone creates more than one in the same action, so we decided to not expose it through the public API to keep things simple. I think you can indeed revert to the P implementation using exceptions before this commit: https://github.com/p-org/P/commit/a3fd5b4b8e8a02861af3bc35f43281d3627a73e8

ankushdesai:
In that case, I will wait for things to get stable and then merge these changes into the P repo.

pdeligia:
Sounds cool, we are now very close to `1.0.0` (without `rc`), most likely the public API is now already stabilized, but we will let you know once `1.0.0` is out to be sure!

ankushdesai:
Thanks @pdeligia, looking forward to a PR with `1.0.0`.

ankushdesai:
@pdeligia @lovettchris Hey, would it be possible for you to please update this pull request or help out with it?

lovettchris:
Hi Ankush, how about I meet you half way?  I updated the PR so it compiles with the latest Coyote 1.0.2 bits, but you'll have to use that info to update your code generator so the unit tests pass.  I'm not too familiar with your code generator, so I don't want to get it wrong.

ankushdesai:
> Hi Ankush, how about I meet you half way? I updated the PR so it compiles with the latest Coyote 1.0.2 bits, but you'll have to use that info to update your code generator so the unit tests pass. I'm not too familiar with your code generator, so I don't want to get it wrong.

This is very helpful! Thanks a lot. I will fix the generator.

ankushdesai:
@pdeligia @lovettchris I was porting old changes to handle early termination of functions in P by throwing exceptions and I noticed that Coyote no longer supports:

```
public override bool GetExceptionHandledLog(ActorId id, string stateName, string actionName, Exception ex, out string text)

        public override bool GetExceptionThrownLog(ActorId id, string stateName, string actionName, Exception ex, out string text)
``` 

How do I make sure that certain exceptions that are purposely thrown done show-up in the logs?

pdeligia:
Hey @ankushdesai, these 2 log APIs are still there, just they got renamed:
```c#
void OnExceptionThrown(ActorId id, string stateName, string actionName, Exception ex);
void OnExceptionHandled(ActorId id, string stateName, string actionName, Exception ex);
```
Just need to override these to change the behavior!

ankushdesai:
> Hey @ankushdesai, these 2 log APIs are still there, just they got renamed:
> 
> ```cs
> void OnExceptionThrown(ActorId id, string stateName, string actionName, Exception ex);
> void OnExceptionHandled(ActorId id, string stateName, string actionName, Exception ex);
> ```
> 
> Just need to override these to change the behavior!

Thanks!!

ankushdesai:
Updated P to use Coyote 1.0.3! @pdeligia @lovettchris Thanks!

pdeligia:
Awesome news @ankushdesai!!

PR #251, opened 2020-03-04
Added valid CMake for ForeignTypes example

ShivKushwah:


Src/Samples/ForeignTypes/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+set(projectName ForeignTypes)
+find_package (Threads)
+add_definitions(-DPRT_PLAT_LINUXUSER)
+include_directories(~/Documents/P/Bld/Drops/Prt/include)
+add_executable(foreign_types
+	ForeignStringType.c
+	ForeignStringType.h
+	ForeignTypesImpl.c
+	Main.c
+	${CMAKE_SOURCE_DIR}/../../../../P/Bld/Drops/Prt/include/ext_compat.h
```
ankushdesai: Instead of adding these files, can we do something like add_directory?

ShivKushwah:
Made the necessary changes @ankushdesai !

PR #259, opened 2020-06-01
Added the override for the Equals methods.

ankushdesai:


ankushdesai:
> we're up to 1.0.11 now and have split systematic testing out to a new nuget package Microsoft.Coyote.Test.

Are there some major code changes after this split? Or I just need to import two packages now thats it?

Thanks!

lovettchris:
We keep a summary of changes made in this [history.md file](https://github.com/microsoft/coyote/blob/master/History.md).

PR #261, opened 2020-07-27
CompilerCore build: Use "java" in $PATH instead of hard-coding to /usr/bin/java

KelvinLi2020:
Not all users install Java into /usr/bin/java, but we can expect it to be available in the PATH variable.


ankushdesai:
I have not come across anyone complaining about this. Hence, I would not like to change it and closing the issue for now.

PR #268, opened 2020-10-30
Add P icon and project metadata to settings

zeynepsu:
This will enable us to publish on nuget.
Version is hard coded for now, will be controlled by a workflow soon.


ankushdesai (changes requested):
This change only packs the P compiler into the package. Lets create a separate package for CSharpRuntime so that programmers can just include it as nuget package to get access to the runtime when compiling the generated C# code. Thanks!

ankushdesai (changes requested):
This is great @zeynepsu , one final request, can you please update the `csproj` files in the Tutorials folder to use these nuget packages instead of hard-coded project paths as references.

Thanks!

ankushdesai (approved):
Thanks @zeynepsu. This looks great!
It would be great to auto upload the packages!

PR #274, opened 2020-11-07
Separate build and test Github workflows

zeynepsu:
Separate each OS into its own Github build and test workflow.


ankushdesai (approved):
Thanks!

ankushdesai (changes requested):
Can you please change the name of each workflow to something like: `CI on Windows`, CI on Mac` and `CI on Ubuntu`.

Also, add badges for these workflows in the ReadMe.

ankushdesai (approved):

PR #288, opened 2020-11-23
Implement java runtime lib for P datatypes.

yzhang90:


ankushdesai (changes requested):
I had two minor cosmetic changes:
1) rename the package to `p.runtime.values`
2) move all the implementation of values into a folder `values`
3) replace all `Prt` to `P`, for example lets call all `PrtSet` as `PSet`

ankushdesai (changes requested):
Would it be possible to build this runtime into a java library or module?
And then reference it along with rvm library for monitoring the implementation?

yzhang90:
> Would it be possible to build this runtime into a java library or module?
> And then reference it along with rvm library for monitoring the implementation?

Do you mean build them as a jar file?

virgil-serbanuta:
> Would it be possible to build this runtime into a java library or module?
> And then reference it along with rvm library for monitoring the implementation?

The answer is, most likely, "yes", but I'm not sure what you mean.

* Do you mean that we should use a different package, e.g. com.runtimeverification.p.runtime? The word "module" seems to suggest that, but the rest of the question seems to suggest that this is not the right interpretation.
* Do you mean that we should package the runtime in a jar? The word "library" seems to suggest that. If so, then there are several ways of doing it:
+ Do you mean that the P repository should only contain a jar file, which would contain only the compiled code? If yes, then we would probably need a separate repository for the source code.
+ Do you mean that the P repository should contain a jar with both the compiled code and the source code? If yes, then we could solve the problem without a separate repository, but it would make it a bit cumbersome to develop/fix stuff.
+ Do you mean that we should have the runtime source code as we have it right now, but that we should have a separate build step in which we pack it into a jar? If so, we could either add the jar to the repository, or we could compile+use it directly.
+ Do you mean that we should move the runtime to a new repository, and have a compilation step which produces a jar, which is then used by our tests/whatever?
+ Any other combination of the above?

virgil-serbanuta:
> I had two minor cosmetic changes:
> 
> 1. rename the package to `p.runtime.values`

FWIW, Java packages should start with domain names in reverse. Choosing which domain to use depends on who owns the code. E.g. if Microsoft "owns" P, then, since Microsoft also owns the microsoft.com domain, then we can use the com.microsoft.p.runtime package. See https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html

That's just a convention, but it's a useful one since it makes sure that, if there are any clashes between package names, the same company owns both packages, so it can solve the issue.

ankushdesai:
> > I had two minor cosmetic changes:
> > 
> > 1. rename the package to `p.runtime.values`
> 
> FWIW, Java packages should start with domain names in reverse. Choosing which domain to use depends on who owns the code. E.g. if Microsoft "owns" P, then, since Microsoft also owns the microsoft.com domain, then we can use the com.microsoft.p.runtime package. See https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html
> 
> That's just a convention, but it's a useful one since it makes sure that, if there are any clashes between package names, the same company owns both packages, so it can solve the issue.

hey, thanks! But P is not owned by Microsoft or Amazon, so would like to avoid this convention. But I do agree with you.

ankushdesai:
> > Would it be possible to build this runtime into a java library or module?
> > And then reference it along with rvm library for monitoring the implementation?
> 
> Do you mean build them as a jar file?

Yes.

ankushdesai:
> > Would it be possible to build this runtime into a java library or module?
> > And then reference it along with rvm library for monitoring the implementation?
> 
> The answer is, most likely, "yes", but I'm not sure what you mean.
> 
> * Do you mean that we should use a different package, e.g. com.runtimeverification.p.runtime? The word "module" seems to suggest that, but the rest of the question seems to suggest that this is not the right interpretation.
> * Do you mean that we should package the runtime in a jar? The word "library" seems to suggest that. If so, then there are several ways of doing it:
> 
> * Do you mean that the P repository should only contain a jar file, which would contain only the compiled code? If yes, then we would probably need a separate repository for the source code.
> * Do you mean that the P repository should contain a jar with both the compiled code and the source code? If yes, then we could solve the problem without a separate repository, but it would make it a bit cumbersome to develop/fix stuff.
> * Do you mean that we should have the runtime source code as we have it right now, but that we should have a separate build step in which we pack it into a jar? If so, we could either add the jar to the repository, or we could compile+use it directly.
> * Do you mean that we should move the runtime to a new repository, and have a compilation step which produces a jar, which is then used by our tests/whatever?
> * Any other combination of the above?

Apologies for the confusion as I am not very used to the java terminology, what I meant was: 
1) Build the runtime into a jar and put it into the Drops folder each time P is built.
2) Use this jar file for compilation of the generated code and running the regressions.
3) No need to create a separate repository, I would prefer keeping all the P code (both compiler and runtime into a single repo).

Hope this helps! Thanks!

virgil-serbanuta:
> > > Would it be possible to build this runtime into a java library or module?
> > > And then reference it along with rvm library for monitoring the implementation?
> > 
> > 
> > The answer is, most likely, "yes", but I'm not sure what you mean.
> > 
> > * Do you mean that we should use a different package, e.g. com.runtimeverification.p.runtime? The word "module" seems to suggest that, but the rest of the question seems to suggest that this is not the right interpretation.
> > * Do you mean that we should package the runtime in a jar? The word "library" seems to suggest that. If so, then there are several ways of doing it:
> > * Do you mean that the P repository should only contain a jar file, which would contain only the compiled code? If yes, then we would probably need a separate repository for the source code.
> > * Do you mean that the P repository should contain a jar with both the compiled code and the source code? If yes, then we could solve the problem without a separate repository, but it would make it a bit cumbersome to develop/fix stuff.
> > * Do you mean that we should have the runtime source code as we have it right now, but that we should have a separate build step in which we pack it into a jar? If so, we could either add the jar to the repository, or we could compile+use it directly.
> > * Do you mean that we should move the runtime to a new repository, and have a compilation step which produces a jar, which is then used by our tests/whatever?
> > * Any other combination of the above?
> 
> Apologies for the confusion as I am not very used to the java terminology, what I meant was:
> 
> 1. Build the runtime into a jar and put it into the Drops folder each time P is built.
> 2. Use this jar file for compilation of the generated code and running the regressions.
> 3. No need to create a separate repository, I would prefer keeping all the P code (both compiler and runtime into a single repo).
> 
> Hope this helps! Thanks!

Thanks, it does help!

I see two options of doing this - we could change the `Bld/build.sh` file to also build this jar. This should be easy enough, but it would not be portable. The other option would be to convince `dotnet build` to run the command to build the jar for us, but last time I looked into this I couldn't figure out how to do it. I could look into this again, but it may take a while, so I thought I'd ask for your opinion first.

ankushdesai:
> > > > Would it be possible to build this runtime into a java library or module?
> > > > And then reference it along with rvm library for monitoring the implementation?
> > > 
> > > 
> > > The answer is, most likely, "yes", but I'm not sure what you mean.
> > > 
> > > * Do you mean that we should use a different package, e.g. com.runtimeverification.p.runtime? The word "module" seems to suggest that, but the rest of the question seems to suggest that this is not the right interpretation.
> > > * Do you mean that we should package the runtime in a jar? The word "library" seems to suggest that. If so, then there are several ways of doing it:
> > > * Do you mean that the P repository should only contain a jar file, which would contain only the compiled code? If yes, then we would probably need a separate repository for the source code.
> > > * Do you mean that the P repository should contain a jar with both the compiled code and the source code? If yes, then we could solve the problem without a separate repository, but it would make it a bit cumbersome to develop/fix stuff.
> > > * Do you mean that we should have the runtime source code as we have it right now, but that we should have a separate build step in which we pack it into a jar? If so, we could either add the jar to the repository, or we could compile+use it directly.
> > > * Do you mean that we should move the runtime to a new repository, and have a compilation step which produces a jar, which is then used by our tests/whatever?
> > > * Any other combination of the above?
> > 
> > 
> > Apologies for the confusion as I am not very used to the java terminology, what I meant was:
> > 
> > 1. Build the runtime into a jar and put it into the Drops folder each time P is built.
> > 2. Use this jar file for compilation of the generated code and running the regressions.
> > 3. No need to create a separate repository, I would prefer keeping all the P code (both compiler and runtime into a single repo).
> > 
> > Hope this helps! Thanks!
> 
> Thanks, it does help!
> 
> I see two options of doing this - we could change the `Bld/build.sh` file to also build this jar. This should be easy enough, but it would not be portable. The other option would be to convince `dotnet build` to run the command to build the jar for us, but last time I looked into this I couldn't figure out how to do it. I could look into this again, but it may take a while, so I thought I'd ask for your opinion first.

How about making the runtime project a maven project and compiling it through build.sh. 
I didn't completely understand, why it wont be cross platform. For windows we can do build.ps1?

virgil-serbanuta:
> > 
> > Thanks, it does help!
> > I see two options of doing this - we could change the `Bld/build.sh` file to also build this jar. This should be easy enough, but it would not be portable. The other option would be to convince `dotnet build` to run the command to build the jar for us, but last time I looked into this I couldn't figure out how to do it. I could look into this again, but it may take a while, so I thought I'd ask for your opinion first.
> 
> How about making the runtime project a maven project and compiling it through build.sh.
> I didn't completely understand, why it wont be cross platform. For windows we can do build.ps1?

Oh, I didn't notice that the project also had a Windows script, thanks.

virgil-serbanuta:
> 
> How about making the runtime project a maven project and compiling it through build.sh.
> I didn't completely understand, why it wont be cross platform. For windows we can do build.ps1?

Done.

Is there a good `com.name` or `org.name` prefix that I could use for the packages?

virgil-serbanuta:
I added one more change, to install the runtime package in the local Maven repository.

Bld/build.ps1
```diff
@@ -6,6 +6,13 @@ git submodule update --init --recursive
 
 # Run the build! :D
 dotnet build -c Release
+
+pushd Src/PRuntimes/RvmRuntime
+mvn install
+popd
+
+Copy-Item -Path "Src/PRuntimes/RvmRuntime/target/*.jar" -Destination "Bld/Drops"
```
ankushdesai, line 14: Can we please put it in `Bld/Drops/RVM`? for all platforms.

ankushdesai (dismissed):
Please address the one comment, I am approving the request otherwise.

Bld/build.ps1
yzhang90: I updated the both build scripts. But I only test the build.sh. I don't have a windows machine to test the build.ps1. please double check the script.

ankushdesai: We can check that the scripts work on windows when we add Github actions for running the RVM regressions.

ankushdesai (approved):
Approved the request. When you create the pull request for adding regressions, I would request you to add a github action to run the RVM regression and also a corresponding separate badge.

PR #299, opened 2020-12-06
fix pproj path extraction on unix case sensitive file systems

sitano:
`option` does not require to preliminary be lower cased.
Case sensitivy is further detected by the IsCorrectPProjFileName.

Without this patch any call on unix cs fs does not work:

    $ ./Bld/Drops/Release/Binaries/netcoreapp3.1/P -proj:./Tutorial/TwoPhaseCommit/TwoPhaseCommit.pproj


ankushdesai (approved):
Thanks for this change. We assumed for this part that the file system is case insensitive.

PR #308, opened 2021-01-26
Updated the code to not create different namespaces

ankushdesai:


ankushdesai:
This is a simple change to remove the namespace names and have a common namespace for the generated code.

PR #309, opened 2021-01-27
Implement ignore keyword

yzhang90:


ankushdesai:
Thanks a lot Yi for these changes!!

PR #313, opened 2021-03-02
Change input AspectJ code, Add aspectj output dir option, Generate BaseAspect

mylibero:
1. Instead of generating AspectJ template file, it outputs BaseAspect that a user does not need to modify
2. P compiler takes "-aspectOutputDir" argument to output all aspectj files into this directory
3. BaseAspect file implements the `@MonitorOn` annotation and short aliases of event handler calls

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/Rvm/AjFileGenerator.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Rvm/AjFileGenerator.cs
@@ -8,7 +8,7 @@ using System.IO;
 
 namespace Plang.Compiler.Backend.Rvm
 {
-    // AjFileGenerator generates the template AspectJ file.
+    // AjFileGenerator generates the base AspectJ file that handles MonitorOn annotations and common pointcuts
     internal class AjFileGenerator
     {
         private CompilationContext Context {get;}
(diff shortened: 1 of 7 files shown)


Src/PCompiler/CompilerCore/Backend/Rvm/AjFileGenerator.cs
```diff
@@ -70,47 +111,72 @@ public IEnumerable<CompiledFile> GenerateSources(Scope globalScope)
...
-            Context.WriteLine(output, "}");
-            Context.WriteLine(output);
-        }
+            Context.WriteLine(output, "pointcut CommonPointCut() :");
+            Context.WriteLine(output, "    !within(com.runtimeverification.rvmonitor.java.rt.RVMObject+) &&");
+            Context.WriteLine(output, "    !adviceexecution() &&");
+            Context.WriteLine(output, "    notwithin();");
+            Context.WriteLine(output, "pointcut UnitTestCase() : ");
+            Context.WriteLine(output, "    execution(* junit.framework.TestCase+.test*(..)) ||");
+            Context.WriteLine(output, "    execution(* s3.junit.S3TestCase+.test*(..)) ||");
```
ankushdesai, line 120: Can we rename s3 to something else? We should not put anything specific to S3 in the P repo.

ankushdesai (changes requested):

Src/PCompiler/CompilerCore/Backend/Rvm/AjFileGenerator.cs
mylibero: I can remove this entirely.

mylibero:
@ankushdesai Please review this again. The hard-code issue is gone now.

ankushdesai (approved):
Thanks looks good.

PR #315, opened 2021-03-03
Fix Rvm RegressionTests to use new input file interface

mylibero:


Tst/RegressionTests/Rvm/run_all_unittests.py
```diff
@@ -34,7 +34,7 @@ def buildCommand(script_directory, temporary_directory, test_name):
 
 def main(argv):
   if len(argv) == 0:
-    parallelism = 3 * tools.findAvailableCpus() / 2
+    parallelism = tools.findAvailableCpus() / 2
```
mylibero, line 37: This is the main fix for the deadlock issue.
The problem was that there were too many processes and threads passing stdout to the parent process and the grandparent process at the same time.
Python2.7 has some structural issues in the subprocess module and I could not find any good solution than reducing parallelism.
Moving to python3.7+ could be a good solution, but it requires porting all python codes to python3 and change to the CI machines.

ankushdesai:
Thanks a lot @mylibero for this fix!

PR #323, opened 2021-04-08
Fix a minor bug, Add .net core dependency in compilercore for creating a linux binary

mylibero:


Src/PCompiler/CompilerCore/CompilerCore.csproj
```diff
@@ -11,6 +11,7 @@
     <PackageReference Include="Antlr4" Version="4.6.5-beta001" />
     <PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
     <PackageReference Include="System.ValueTuple" Version="4.5.0" />
+    <PackageReference Include="Microsoft.NETCore.Targets" Version="3.1.0" PrivateAssets="all" />
```
ankushdesai, line 14: Can you please tell me what this change does? and why is it needed?

mylibero: It explicitly states dependency to dotnetcore 3.1. I need it for packaging this into a standalone linux executable that does not require dotnet to run.
Without this, dotnet complains about ambiguous dependencies.

ankushdesai (approved):

PR #372, opened 2021-10-14
Fix typos in the Language Manual

claudiacauli:
Spotted these while reading, hope it helps :)

- In P top-level grammar, changed 'TypeDefDecl' to 'TypeDecl' to be consistent
with the syntax given for the type declarations in datatypes.md
- In description of type declarations Module System, changed 'them' to 'it' as
it seems to be that the pronoun refers to "system".
- 'helps' -> 'help'
- In datatypes.md, 'cannot to assigned to' -> 'cannot be...'
- In monitors.md, 'it asserts' -> 'assert' as it seems to me that the sentence
is talking about 'spec machines'; and few more similar typos.
- In statements.md, b should print ["b"] (not "a").
- And a few more similar ones :)


ankushdesai:
Thanks a lot Claudia!

PR #438, opened 2022-06-06
[PSymbolicRuntime] Bump dependency versions

dijkstracula:
Swap out Log4J 1.x dependency with the Log4j1-2 adaptor instead and upgrade guava to 30.0.

Tested with the `./build.sh` script in `Src/PRuntimes/PSymbolicRuntime/scripts`.


ankushdesai:
Thanks a lot!

PR #439, opened 2022-06-07
[TypeChecker] Accumulate all tuple types into Scope

dijkstracula:
This is part of the non-Java compiler work required to implement https://github.com/dijkstracula/prtsandbox/issues/12 .

Ankush, you might not like my choice of implementing `TupleType::ToNamedTupleType()`.  Let me know and we can iterate on that.


ankushdesai:
Looks good to me! thanks!

PR #440, opened 2022-06-08
[TypeChecker] After TypeDef type resolution, stash tuples' name

dijkstracula:
This is needed to improve the ergonomics of generated payload classes.

I'm not sure DeclarationVisitor is the right place for this; it almost feels like it should live in a separate Visitor that walks the AST after the current DeclarationVisitor.  What do you think?


dijkstracula:
Noticed a bug.  Will fix and resubmit.

PR #445, opened 2022-06-13
[Java] emit right fn signature for handler of no arguments

dijkstracula:
This is an issue that I realised when working on https://github.com/p-org/P/pull/444 earlier today:

Anonymous event handlers need not consume the event, such as here:

```
     1  type t = (a : seq[int]);
     2  event ev : t;
     3
     4  event ev2 : int;
     5
     6  spec foo observes ev, ev2 {
     7    start state Init {
     8      on ev, ev2 do {
     9
    10      }
    11    }
    12
    13  }
```

Even though `ev` and `ev2` have incompatable payload types, this code is still
safe since we never do anything with the payload.  However, the current code generator
would incorrectly emit a parameter to the generated method, which would cause
the generated Java code to fail typechecking, like so.

```
 58         private void Anon(ev pEvent) {
 59
 60         }
 ...
 64             addState(new State.Builder(INIT_STATE)
 65                 .isInitialState(true)
 66                 .withEvent(ev.class, this::Anon)
 67                 .withEvent(ev2.class, this::Anon) // <-- This fails to typecheck!
 68                 .build());
```

This patch addresses cases where the Function signature's parameter list is
empty, and does not emit an argument for it.  At monitor construction time, the
event handler for `ev` and `ev2` both consume the event, as is required for
typechecking to succeed, but discards the result before calling the actual
handler.

```
 58         private void Anon() {
 59
 60         }
 ...
 64             addState(new State.Builder(INIT_STATE)
 65                 .isInitialState(true)
 66                 .withEvent(ev.class, __ -> Anon(); )
 67                 .withEvent(ev2.class, __ -> Anon(); )
 68                 .build());
```

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/Java/JavaCodeGenerator.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/JavaCodeGenerator.cs
@@ -408,7 +408,20 @@ namespace Plang.Compiler.Backend.Java {
             string args;
             if (f.IsAnon)
             {
-                args = $"{_eventArgForFn[f].Name} pEvent";
+                if (f.Signature.Parameters.Count == 0)
+                {
+                    args = "";
+                }
+                else if (f.Signature.Parameters.Count == 1)
+                {
+                    args = $"{_eventArgForFn[f].Name} pEvent";
+                }
+                else
+                {
+                    int line = f.SourceLocation.start.Line;
+                    throw new Exception(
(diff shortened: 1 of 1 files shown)


ankushdesai (approved):
Thanks! looks cool!

ankushdesai:
@dijkstracula please resolve the conflict.

dijkstracula:
@ankushdesai done, thanks.

Issue #447, opened 2022-06-15
Unescaped strings in trace output leads to ambiguity


dijkstracula:
Currently, when we run `pmc` with `-v`, event payloads are dumped out to the trace file just by the value produced by each `PrtValue`'s `.ToString()` method.  There are some implications of this if we want to read these trace files back in for replaying versus just being for human eyeballing:  We do not escape strings or values contained within the strings.  This is problematic.  Consider the following P program:

```
nathta@bcd0741cf59d /tmp % cat -n foo.p
     1  event ev : (s:string);
     2  machine Foo {
     3    start state Init { }
     4  }
     5
     6  machine Driver {
     7    var f: Foo;
     8
     9    start state Init {
    10      entry {
    11        f = new Foo();
    12        send f, ev, (s = "a:string, with:problems",);
    13      }
    14    }
    15  }
    16
    17  test doIt [main=Driver]: (union {Driver}, {Foo});
```

Notice the string value to the named tuple on line 12: it contains colons (which are used to separate keys and values in a named tuple) and `, `, which separates key-value pairs in a named tuple's output.  When we try to run this program, we get:

```
nathta@bcd0741cf59d /tmp % coyote test POutput/netcoreapp3.1/foo.dll -v
. Testing POutput/netcoreapp3.1/foo.dll
...
<SendLog> 'PImplementation.Driver(2)' in state 'Init_1' sent event 'ev with payload (<s:a:string, with:problems, >)' to 'Foo(3)'.
...
```

There's no way for the parser to disambiguate `<s:a:string, with:problems, >` as a `(s: string, with: string)` or `(s: string)`.  

One solution could be to modify `PrtString.ToString()` to wrap the value in double quotes, and then escape `,` `[`, `]`, `<`, `>`, `:`, `"` and `'` metacharacters (and possibly others, like `\` if we're using that character as the escape character ?).  This means the parser will, of course, need to check for escaped values but then at least things should in theory less ambiguous.

Another solution would be to contemplate a different data structure representation that's distinct from each PrtValue's `ToString()` output since I feel as though there will be a million corner cases here to stomp out.

dijkstracula:
Update: modifying `PrtString#toString` seems like a brittle solution as `dotnet test -c Release` starts to fail.

dijkstracula:
We did the simple thing in https://github.com/p-org/P/pull/448 ; if that is insufficient we can reopen this issue.

PR #448, opened 2022-06-16
[PCSharpRuntime] Log an escaped string representation of PrtValues

dijkstracula:
This is an attempt to address https://github.com/p-org/P/issues/447 , but we
should think carefully about whether we've covered our bases on this.

Previously, the PLogFormatter would emit the direct string representation
of any PrtValues in an event payload.  This means that there could be ambiguity
owing to PrtStrings not being escaped.  For instance, the P statement

```
send f, ev, (s = "a:string, with:\"problems",);
```

Would yield a log line of the form:

```
<SendLog> 'PImplementation.Driver(2)' in state 'Init_1' sent event 'ev with payload (<s:a:string, with:"problems, >)' to 'Foo(3)'
```

Which is ambiguous in terms of what the NamedTuple arity is, and also
results in an unclosed string literal.

This patch introduces a variation on ToString(), called ToEscapedString(),
which ensures that problematic metacharacters are quoted, and PrtStrings
are themselves enclosed in double-quotes.  The above log line would now be

```
<SendLog> 'PImplementation.Driver(2)' in state 'Init_1' sent event 'ev with payload (<s:"a:string, with:\"problems", >)' to 'Foo(3)'.
```

This change could not be done as a modification of ToString(), since the P
runtime relies on a "naked" representation of the value.

### DIFF
--- a/Src/PRuntimes/PCSharpRuntime/PLogFormatter.cs
+++ b/Src/PRuntimes/PCSharpRuntime/PLogFormatter.cs
@@ -28,7 +28,7 @@ namespace Plang.CSharpRuntime
             }
             else
             {
-                var withPayload = ((PEvent)e).Payload == null ? "" : $" with payload ({((PEvent)e).Payload})";
+                var withPayload = ((PEvent)e).Payload == null ? "" : $" with payload ({((PEvent)e).Payload.ToEscapedString()})";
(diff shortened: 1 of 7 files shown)


ankushdesai (approved):
Great

PR #449, opened 2022-06-17
Nathan/entry handler args

dijkstracula:
Event handler functions consume a subclass of PEvent and then in the
    function body the payload is extracted.  However, non-nullary entry
    and exit handlers don't consume a PEvent but rather take a "payload"
    directly, so some infrastructure needed to be built up to accommodate
    those functions.

Following that, we now emit entry/exit handler function wiring (oops!)

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/Java/JavaCodeGenerator.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/JavaCodeGenerator.cs
@@ -19,13 +19,6 @@ namespace Plang.Compiler.Backend.Java {
         private CompiledFile _source;
         private Scope _globalScope;
 
-        // In the C# and Rvm compilers, event handlers' signatures are hard-coded to
-        // be an abstract Event type.  However, our Java runtime takes advantage of the
-        // typechecker to be able to specify exactly _which_ event it is going to be
-        // handed.  This isn't something explicitly handed to us in the AST so we
-        // accumulate it ourselves before proceeding with code generation.
-        private Dictionary<Function, PEvent> _pEventForEventHandler;
-
         /// <summary>
         /// Generates Java code for a given compilation job.
         ///
@@ -41,8 +34,6 @@ namespace Plang.Compiler.Backend.Java {
             _context = new CompilationContext(job);
             _source = new CompiledFile(_context.FileName);
             _globalScope = scope;
-
-            _pEventForEventHandler = CollateEventHandlerToEventMappings();
 
             WriteImports();
             WriteLine();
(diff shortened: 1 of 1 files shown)


Src/PCompiler/CompilerCore/Backend/Java/JavaCodeGenerator.cs
```diff
@@ -399,13 +399,38 @@ private void WriteFunctionSignature(Function f)
                 }
                 else if (f.Signature.Parameters.Count == 1)
                 {
-                    args = $"{_eventArgForFn[f].Name} pEvent";
+                    // Two possibilities:
```
ankushdesai, line 402: I dont understand this distinction. We can have a generic interface. All event handlers be it entry or transition take as input the event. You just have to introduce a special `constructor` event which is sent when the machine or monitor gets created and also a `goto` event if you enter the state on a goto statement. For transitions `on E1 goto S1` the entry function will receive the event E1.

ankushdesai: goto and contructor are like `internal` events and just make sure the event names dont conflict with user defined events

ankushdesai (changes requested):

Src/PCompiler/CompilerCore/Backend/Java/JavaCodeGenerator.cs
dijkstracula: > All event handlers be it entry or transition take as input the event.

I think there might be different definitions of "event" floating around, so let me take a minute and walk through what's going on here.   From my view there is a PEvent that has a distinct payload.  The function signature for a function designates the _payload_ of the event, not the PEvent itself, which is why I have this layer of indirection through `_eventArgForFn` - I need to know the PEvent associated with a handler so I can emit the correct function signature.

Concretely, consider the following code:

```
  1 type tRequest = (rId: int);
  2 event eRequest : tRequest;
  3
  4 spec ReqIdsAreMonotonicallyIncreasing observes eRequest {
  5     // keep track of the Id in the previous request
  6     var previousId : int;
  7     start state Init {
  8         on eRequest do (req: tRequest) {
  9             assert req.rId > previousId,
 10             format ("Request Ids not monotonically increasing, got {0},
 11             previously seen Id was {1}", req.rId, previousId);
 12             previousId = req.rId;
 13         }
 14     }
 15 }
```
The P function signature consumes not an `eRequest` event but the `tRequest` payload.

In both the C# and Java implementations, we end up with code that looks like this:

```java
 19     public static class PTuple_rId implements Values.PTuple<PTuple_rId> {
 20         public int rId;
 ...
 59     public record eRequest(PTuple_rId payload) implements PObserveEvent.PEvent { }
...
 68         private void Anon(eRequest pEvent) {
 69             PTuple_rId req = pEvent.payload;
...
 78             TMP_tmp0 = req.rId;
 79             TMP_tmp1 = TMP_tmp0 > previousId;
```

Correspondingly, at that line, `f`'s parameter type is a TypeDef whose canonical representation is `"(rId:int)"`.  If we emitted this type directly, the generated Java code would look like this:

```java
 68         private void Anon(PTuple_rId pEvent) {
 69             // of course, we don't need PTuple_rId req = pEvent.payload; any longer.
 70             int TMP_tmp0 = 0;
 71             boolean TMP_tmp1 = false;
```

I did it this way to maintain the same signatures as the C# generated code:

```
 38         public void Anon(Event currentMachine_dequeuedEvent)
 39         {
 40             ReqIdsAreMonotonicallyIncreasing currentMachine = this;
 41             PrtNamedTuple req = (PrtNamedTuple)(gotoPayload ?? ((PEvent)currentMachine_dequeuedEvent).Payload);
 42             this.gotoPayload = null;
 43             PrtInt TMP_tmp0 = ((PrtInt)0);
```

I'm happy to unify this per the TODO statement below your comment but it felt like there must be a reason why it was done this way (else why wouldn't the C# code simply emit an argument of type `PrtNamedTuple`?).  This will require some changes that I have to dig into anyway, since it relates to emitting records which we can't do anymore anyway.

dijkstracula:
As of https://github.com/p-org/P/pull/449/commits/5c5b602b308d386098be728e7ee836c9bbf56949 , we would now generate:

```java
 68         private void Anon(PTuple_rId req) {
...
 77             TMP_tmp0 = req.rId;
 78             TMP_tmp1 = TMP_tmp0 > previousId;
```

this certainly is a cleaner implementation on the codegen side as well as on the generated code side.  This would require a minor Java runtime change which I'll look into now.

dijkstracula:
Looking at the generated C# code again, I wonder if that indirection happens as an artefact of using Coyote's particualr event-handling API?

dijkstracula:
@ankushdesai I have a patch for the runtime that works correctly with the code emitted on this branch; let me know if this one works for you.  I certainly agree that this is cleaner across the board than what I was doing before!

ankushdesai (approved):
Fantastic! I like this one. Much cleaner

PR #452, opened 2022-06-20
[Java] emit POJOs rather than records for PEvents

dijkstracula:
This now generates plain old java classes for records, which were introduced in Java 14 and are not supported by the version of Java that we need.

Example output:

```
 event eIncrVal : (i : int);
```

```java
    public static class eIncrVal implements PObserveEvent.PEvent<PTuple_i> {
        public eIncrVal(PTuple_i p) { this.payload = p; }
        private PTuple_i payload;
        public PTuple_i getPayload() { return payload; }

        @Override
        public boolean equals(Object o) {
            if (this == o) return true;
            if (o == null || getClass() != o.getClass()) return false;
            eIncrVal that = (eIncrVal)o;
            return Values.deepEquals(this.payload, that.payload);
        }

        @Override
        public int hashCode() { return Objects.hash(payload); }

        @Override
        public String toString() {
            return "eIncrVal[" + payload + "]";
        } // toString()

    } // PEvent definition for eIncrVal
```

Morally this is no different from emitting records; we just have to build up the mechanism ourselves in the class definition.

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/Java/JavaCodeGenerator.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/JavaCodeGenerator.cs
@@ -220,16 +220,38 @@ namespace Plang.Compiler.Backend.Java {
             string eventName = _context.Names.GetNameForDecl(e);
 
             TypeManager.JType argType = _context.Types.JavaTypeFor(e.PayloadType);
-            switch (argType)
+            bool hasPayload = !(argType is TypeManager.JType.JVoid);
+
+            WriteLine($"public static class {eventName} extends PObserveEvent.PEvent<{argType.ReferenceTypeName}> {{");
+
+            if (hasPayload)
             {
-                case TypeManager.JType.JVoid _:
-                    // Special-case an event with no payload: just emit an empty record.
-                    WriteLine($"public record {eventName}() implements PObserveEvent.PEvent {{ }} ");
-                    break;
-                default:
-                    WriteLine($"public record {eventName}({argType.TypeName} payload) implements PObserveEvent.PEvent {{ }} ");
-                    break;
+                WriteLine($"public {eventName}({argType.TypeName} p) {{ this.payload = p; }}");
             }
+            else
+            {
+                WriteLine($"public {eventName}() {{ }}");
+            }
+
+            WriteLine($"private {argType.ReferenceTypeName} payload; ");
+            WriteLine($"public {argType.ReferenceTypeName} getPayload() {{ return payload; }}");
+            WriteLine();
+
+            WriteLine("@Override");
+            WriteLine("public String toString() {");
+            if (hasPayload)
+            {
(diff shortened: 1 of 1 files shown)


Src/PCompiler/CompilerCore/Backend/Java/JavaCodeGenerator.cs
```diff
@@ -220,17 +220,52 @@ private void WriteEventDecl(PEvent e)
...
+            {
+                WriteLine($"public {eventName}() {{ }}");
             }
 
+            WriteLine($"private {argType.ReferenceTypeName} payload; ");
+            WriteLine($"public {argType.ReferenceTypeName} getPayload() {{ return payload; }}");
+            WriteLine();
+
+            WriteLine("@Override");
+            WriteLine("public boolean equals(Object o) {");
```
ankushdesai, line 241: Can this by any chance be moved into a base class? Instead of generating it for each event?

ankushdesai (changes requested):

Src/PCompiler/CompilerCore/Backend/Java/JavaCodeGenerator.cs
ankushdesai: I meant the generic code like `hashCode` and `equals`

dijkstracula: I don't think so - how would the base class look?  We need to do a field-by-field comparison, and the base class would have no way of iterating over those fields (short of something gnarly like using reflection, which is best avoided here I think)

dijkstracula: That said, though, we _could_ simplify `equals` by calling directly into `Values.deepEquals()`, since all those checks are already done on the prt side.

ankushdesai: I like both options, anything that reduces the generated code is great!

dijkstracula: Yep, thanks for this observation!  After [this change](https://github.com/dijkstracula/prtsandbox/commit/4ad482a11fdce75d60f11d463c37afa6d0348cb9) in the runtime, we don't need to emit that boilerplate over and over again.

dijkstracula:
After the above change, the above P event definition now looks thus:

```
 90     public static class eIncrVal extends PObserveEvent.PEvent<PTuple_i> {
 91         public eIncrVal(PTuple_i p) { this.payload = p; }
 92         private PTuple_i payload;
 93         public PTuple_i getPayload() { return payload; }
 94
 95         @Override
 96         public String toString() {
 97             return "eIncrVal[" + payload + "]";
 98         } // toString()
 99
100     } // PEvent definition for eIncrVal
```

ankushdesai:
Great!

PR #453, opened 2022-06-20
Java FFI

dijkstracula:
This is the foreign function interface for P -> Java compilation.

so I have foreign functions working and compiling, I think.

Consider the following P program:

```
nathta@bcd0741cf59d /tmp % cat -n foo.p
     1  event eIncr;
     2
     3  // A foreign type
     4  type Counter;
     5
     6  // Foreign functions that operate on a foreign type
     7  fun NewCounter(): Counter;
     8  fun GetNextCount(counter: Counter): int;
     9
    10  spec foo observes eIncr {
    11    var i : int;
    12    var c : Counter;
    13
    14    start state Init {
    15      entry {
    16        i = 0;
    17        c = NewCounter();
    18      }
    19
    20      on eIncr do {
    21        i = i + 1;
    22        assert i == GetNextCount(c), "uh oh";
    23      }
    24    }
    25  }
```

Nothing too terribly interesting.  In the way similar to the C# compilation path, programmers will define the Counter class:

```
nathta@bcd0741cf59d /tmp % cat -n Counter.java
     1  public class Counter {
     2      public int i;
     3      public Counter() { i = 0; }
     4  }
```

 And then a set of static methods to act as foreign functions.  Because we. don’t have partial classes in Java, we have to create a proxy class to bridge them together:

```
nathta@bcd0741cf59d /tmp % cat -n foo_FFIBridge.java
     1
     2  public class foo_FFIBridge {
     3      public static Counter NewCounter() {
     4          return new Counter();
     5      }
     6
     7      public static int GetNextCount(Counter c) {
     8          c.i += 1;
     9          return c.i;
    10      }
    11  }
```

The naming convention is that foreign functions for XYZ should go in a file called XYZ_FFIBridge.java in lieu of implementing a partial class named XYZ.  So it goes.

We need a way of generating a `pom.xml` or some such build script to compile everything, which I'll put out as a separate ticket.  In the mean time, we can compile things manually with `javac`. Of course, this is not the most ergonomic way to compile and we wouldn’t ask customers to do this.

```

nathta@bcd0741cf59d /tmp % pcl -generate:java foo.p
Parsing ..
Type checking ...
Code generation ....
Generated foo.java
nathta@bcd0741cf59d /tmp % javac -source 11 -cp ~/IdeaProjects/EventExperiments/target/EventExperiments-1.0-SNAPSHOT.jar foo.java Counter.java foo_FFIBridge.java
warning: [options] system modules path not set in conjunction with -source 11
1 warning
nathta@bcd0741cf59d /tmp % jar -cf foo.jar foo.class Counter.class foo_FFIBridge.class
nathta@bcd0741cf59d /tmp % jar -tf foo.jar
META-INF/
META-INF/MANIFEST.MF
foo.class
Counter.class
foo_FFIBridge.class
nathta@bcd0741cf59d /tmp %
```

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/Java/NameManager.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/NameManager.cs
@@ -1,5 +1,7 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using Plang.Compiler.TypeChecker.AST;
 using Plang.Compiler.TypeChecker.AST.Declarations;
 using Plang.Compiler.TypeChecker.AST.States;
@@ -34,19 +36,53 @@ namespace Plang.Compiler.Backend.Java
         /// <returns>A consistent but distinct identifier for `t`.</returns>
         internal string NameForNamedTuple(NamedTupleType t)
         {
-            string val;
-            if (!_namedTupleJTypes.TryGetValue(t, out val))
+            if (!_namedTupleJTypes.TryGetValue(t, out var val))
             {
                 IEnumerable<string> names = t.Names;
-                if (names.Count() > 5)
-                {
-                    names = names.Select(f => f.Substring(0, 3)).ToArray();
-                }
+                names = names.Select(AbbreviateTupleName).ToArray();
                 val = UniquifyName("PTuple_" + string.Join("_", names));
                 _namedTupleJTypes.Add(t, val);
             }
 
             return val;
+        }
+
+        /// <summary>
+        /// Attempts to simplify the given tuple name while still leaving it somewhat readable.
+        /// </summary>
+        /// <param name="name">The token to simplify</param>
+        /// <returns>The simplified token.</returns>
+        private static string AbbreviateTupleName(string name)
+        {
+            // Short names are already as succinct as we can reasonably expect.
+            if (name.Length <= 5)
+            {
+                return name;
+            }
+
+            // Is it a superLongCamelCaseFieldName?  Abbreviate it that way.
+            if (name.Count(char.IsUpper) >= 3)
+            {
+                return String.Concat(name.Where(char.IsUpper)).ToLower();
+            }
+
+            // Strip and simplify some common prefixes and suffixes.
+            name = Regex.Replace(name, "^is([A-Z].*)", "$1");
+            name = Regex.Replace(name, "(.*)Id$", "$1");
+            name = Regex.Replace(name, "(.*)Val$", "$1");
+            if (name.Length <= 5)
+            {
+                return name;
+            }
+
+            // If not, strip the vowels in the middle of a word out so it still seems pronounceable at a distance.
+            name = name.Substring(0, 1) + Regex.Replace(name.Substring(1), "[aeiou]", "");
+            if (name.Length <= 5)
+            {
+                return name.ToLower();
+            }
+
+            return name.Substring(0, 5).ToLower();
         }
 
         protected override string ComputeNameForDecl(IPDecl decl)
--- a/Src/PCompiler/CompilerCore/Backend/Java/NameManager.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/NameManager.cs
@@ -79,5 +79,28 @@ namespace Plang.Compiler.Backend.Java
 
             return UniquifyName(name);
         }
+
+        /// <summary>
+        /// Produces the class name for the foreign function bridge class for a given machine.
+        ///
+        /// In the C# code generator, partial classes are used to automatically weave together
+        /// the machine class and its foreign functions.  For instance, a machine called Client
+        /// that relies on foreign functions would have the latter implemented in a partial class
+        /// also called Client.  The C# compiler would then merge the two together into a final
+        /// class definition.
+        ///
+        /// Java does not have this language feature, unfortunately, so we need a different approach:
+        /// we need foreign function writers to implement foreign functions as static methods
+        /// in a class whose name is derived from the monitor class.  The naming convention is
+        /// specified by the return value of this function.
+        ///
+        /// <see url="https://en.wikipedia.org/wiki/Bridge_pattern"/>
+        /// </summary>
+        /// <param name="machineName"></param>
+        /// <returns></returns>
+        public string FFIBridgeForMachine(string machineName)
+        {
+            return $"{machineName}FFI";
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


Src/PCompiler/CompilerCore/Backend/Java/NameManager.cs
```diff
@@ -77,7 +116,39 @@ protected override string ComputeNameForDecl(IPDecl decl)
...
+        /// Java does not have this language feature, unfortunately, so we need a different approach:
+        /// we need foreign function writers to implement foreign functions as static methods
+        /// in a class whose name is derived from the monitor class.  The naming convention is
+        /// specified by the return value of this function.
+        ///
+        /// <see url="https://en.wikipedia.org/wiki/Bridge_pattern"/>
+        /// </summary>
+        /// <param name="machineName"></param>
+        /// <returns></returns>
+        public string FFIBridgeForMachine(string machineName)
```
ankushdesai, line 149: The foreign functions can be global as well. What do you do then?

ankushdesai: Is there a separate file for `machine` level and global functions?

ankushdesai (changes requested):

ankushdesai (commented):
Can we just assume that all the foreign code for a P project goes inside a package XYZ.foreign? Can you think of a simpler approach, thanks!

Src/PCompiler/CompilerCore/Backend/Java/NameManager.cs
dijkstracula: Good point - I ought to emit a `GlobalFunctions` like the C# compiler does.

dijkstracula:
> Can we just assume that all the foreign code for a P project goes inside a package XYZ.foreign?

Hm, so one potential issue there might be name collisions - if I have two P files and I associate a foreign function with the same name in each of them (or foreign classes with the same name, etc), do we run into issues there when we build the whole project?

ankushdesai:
> > Can we just assume that all the foreign code for a P project goes inside a package XYZ.foreign?
> 
> Hm, so one potential issue there might be name collisions - if I have two P files and I associate a foreign function with the same name in each of them (or foreign classes with the same name, etc), do we run into issues there when we build the whole project?

This is an excellent point and I dont know what the current behavior is if you define two foreign functions with the same name in two different machines. I think we only allow one definition for it :). This problem exists in the C# world and I agree would be good to fix in the Java world.

How about package names XYZ.MachineName for all machine local functions and XYZ.Global for all global foriegn functions?

dijkstracula:
Sounds good!  

> How about package names XYZ.MachineName for all machine local functions and XYZ.Global
> for all global foriegn functions?

So if XYZ is the project name, I don't think we can name it exactly like this - XYZ becomes a [top level class](https://github.com/dijkstracula/prtsandbox/blob/main/src/main/java/tutorialmonitors/ClientServer.java#L14) and not a namespace or something we can otherwise put things in post-hoc.  I'm definitely not opposed to giving them less gnarly names than FFIBridge, though.

For the moment, though, here are the two cases, which I'll push up now.

## Global FFI

```
nathta@bcd0741cf59d /tmp % cat -n foo.p
     1  event eIncr;
     2
     3  // A global foreign type
     4  type Counter;
     5
     6  // Global foreign functions that operate on a foreign type
     7  fun NewCounter(): Counter;
     8  fun GetNextCount(counter: Counter): int;
     9
    10  spec myMonitor observes eIncr {
    11    var i : int;
    12    var c : Counter;
    13
    14    start state Init {
    15      entry {
    16        i = 0;
    17        c = NewCounter();
    18      }
    19
    20      on eIncr do {
    21        i = i + 1;
    22        assert i == GetNextCount(c), "uh oh";
    23      }
    24    }
    25  }
nathta@bcd0741cf59d /tmp %
```

Emits Java code like this:

```
 33         private void Anon() {
...
 37             TMP_tmp0 = GlobalFFIBridge.NewCounter();
...
 39         }
 40         private void Anon_1() {
...
 50             TMP_tmp2 = GlobalFFIBridge.GetNextCount(TMP_tmp1);
...
 54         }
```

## Per-machine FFI

Now, if I move those foreign function defns inside the monitor:

```
nathta@bcd0741cf59d /tmp % cat -n foo.p
     1  event eIncr;
     2
     3  // A global foreign type
     4  type Counter;
     5
     6  spec myMonitor observes eIncr {
     7
     8    // Machine-local foreign functions that operate on a foreign type
     9    fun NewCounter(): Counter;
    10    fun GetNextCount(counter: Counter): int;
    11
    12    var i : int;
    13    var c : Counter;
    14
    15    start state Init {
    16      entry {
    17        i = 0;
    18        c = NewCounter();
    19      }
    20
    21      on eIncr do {
    22        i = i + 1;
    23        assert i == GetNextCount(c), "uh oh";
    24      }
    25    }
    26  }
nathta@bcd0741cf59d /tmp %
```

(I get a compilation error if I try to do the same with the type, so I left that at the top scope)

Then we emit the per-monitor FFI bridge call instead:

```
 35         private void Anon() {
...
 39             TMP_tmp0 = myMonitor_FFIBridge.NewCounter();
...
 41         }
 42         private void Anon_1() {
...
 51             TMP_tmp1 = (Counter)Values.deepClone(c);
 52             TMP_tmp2 = myMonitor_FFIBridge.GetNextCount(TMP_tmp1);
...
 56         }
```

ankushdesai:
> > what I meant by XYZ.MachineName was that I was assuming all your code inside XYZ namespace or package.
> 
> ah yeah got it! Right now everything is all inside one default unnamed package - you're right that we might be able to do something like that once we can see what the code for larger projects looks like (or generating a java project with a `pom.xml`)

If you can add that, it would be great for the next auto compilation part.

dijkstracula:
>  (or generating a java project with a pom.xml)

Actually, if you would like this decided before we merge all this, then I can integrate generating the project file in this branch and keep pushing to this PR.

dijkstracula:
> If you can add that, it would be great for the next auto compilation part.

So one thing that isn't immediately clear to me: in the above example let's say the package name is the same as the P project: so we'd have something like

```java
package foo;
import prt.*
...
public class foo {
  public static class myMonitor {
    ...
  }
}
```

(Remember, we need the top-level `foo` class unless we want to generate one `.java` output file for every class that it contains because of Java's "one file = one top-level class" rule.)  In the proposal where we generate a package name and reference it, we don't _need_ to qualify accessing `GlobalFFI` and `myMonitorFFI` because they're in the same package.  So, what does that get us?  They're already in the same package.

If the package name _isn't_ the same as the P project, what should it be?

ankushdesai:
Looks good!

ankushdesai (approved):

PR #455, opened 2022-06-21
Generate Java projfile; unify Compile() methods across backends

dijkstracula:
There are two main pieces to this patch:

The first is to generate `pom.xml` build files for the Java backend, and then have the P compiler generate a JAR from that pom file.

```
nathta@bcd0741cf59d pprojtest % tree .
.
├── PForeign
│   ├── globals
│   │   └── Counter.java
│   └── myMonitor.java
├── PSpec
│   └── foo.p
├── README.md
└── pprojtest.pproj

3 directories, 5 files
nathta@bcd0741cf59d pprojtest % pcl -proj:pprojtest.pproj
==== Loading project file: pprojtest.pproj
....... includes p file: /Users/nathta/code/pprojtest/PSpec/foo.p
Parsing ..
Type checking ...
Code generation ....
Generated pom.xml
Generated pprojtest.java
Compiling pom.xml ..

nathta@bcd0741cf59d pprojtest % tree .
.
├── PForeign
│   ├── globals
│   │   └── Counter.java
│   └── myMonitor.java
├── PGenerated
│   └── pprojtest.java
├── POutput
│   ├── classes
│   │   ├── PForeign
│   │   │   ├── globals
│   │   │   │   └── Counter.class
│   │   │   └── myMonitor.class
│   │   └── PGenerated
│   │       ├── pprojtest$DefaultEvent.class
│   │       ├── pprojtest$PHalt.class
│   │       ├── pprojtest$eIncr.class
│   │       ├── pprojtest$myMonitor.class
│   │       └── pprojtest.class
│   ├── generated-sources
│   │   └── annotations
│   ├── maven-archiver
│   │   └── pom.properties
│   ├── maven-status
│   │   └── maven-compiler-plugin
│   │       └── compile
│   │           └── default-compile
│   │               ├── createdFiles.lst
│   │               └── inputFiles.lst
│   └── pprojtest-1.0-SNAPSHOT.jar
├── PSpec
│   └── foo.p
├── README.md
├── pom.xml
└── pprojtest.pproj

16 directories, 18 files
nathta@bcd0741cf59d pprojtest %
```

The second addresses a [TODO](https://github.com/p-org/P/blob/1b5502d23c8c383fdacbedde216e5c56cb50666e/Src/PCompiler/CompilerCore/Compiler.cs#L48) that was dependent on the Java backend reaching this stage.  

Previously, special-cased backends would call a static `Compile()` method.  C# does not allow static methods in interfaces, so this now becomes an instance method on the code generator.  Further, now, all relevant backends have a Compile() method to remove the special case logic in Compiler.cs.  (There was a bunch of refactoring propagation that happened here as a result, since there's really no longer a need to distinguish a `FooCodeGenerator` from a `FooCodeCompiler`.)


ankushdesai (approved):
Nice, looks great. thanks!

PR #459, opened 2022-06-23
[Java] generate Java source for events in their own file

dijkstracula:
Previously all Java code was emitted into a single Java file, named after the P project.  This was just like the C# code generator.

This patch now generates Events in their own file, called `Events.java` (but still in the same package).  As a forward reference for future changes, that blob of Java code is now called `Machines.java` - as there are now multiple code generators working in tandem, and since identifiers need to be referenced between files, it proved to be exceedingly awkward to thread through the top level name for this file.

```
nathta@bcd0741cf59d pprojtest % cat -n PSpec/foo.p
     1  type t = (a: int, b: seq[int]);
     2
     3  event eIncr : t;
     4  event eOther;
     5  event eYetAnother;
     6
     7  spec monitorA observes eIncr {
     8    start state Init {
     9      entry {}
    10      on eIncr do {}
    11    }
    12  }
    13
    14  spec monitorB observes eOther {
    15    start state Init {
    16      entry {}
    17      on eOther do {}
    18    }
    19  }
nathta@bcd0741cf59d pprojtest % pcl -proj:pprojtest.pproj
==== Loading project file: pprojtest.pproj
....... includes p file: /Users/nathta/code/pprojtest/PSpec/foo.p
Parsing ...
Type checking ...
Code generation ...
Reusing existing pom.xml
Generated Machines.java.
Generated Events.java.
Compiling pprojtest...
nathta@bcd0741cf59d pprojtest % 
```

The Events file looks thus:

```java
nathta@bcd0741cf59d pprojtest % cat -n PGenerated/Events.java
     1  package PGenerated;
     2  import java.util.*;
     3
     4  /***************************************************************************
     5   * This file was auto-generated on Thursday, 23 June 2022 at 16:45:35.
     6   * Please do not edit manually!
     7   **************************************************************************/
     8
     9
    10  public class Events {
    11      public static class eIncr extends prt.events.PEvent<Machines.PTuple_a_b> {
    12          public eIncr(Machines.PTuple_a_b p) { this.payload = p; }
    13          private Machines.PTuple_a_b payload;
    14          public Machines.PTuple_a_b getPayload() { return payload; }
    15
    16          @Override
    17          public String toString() { return "eIncr[" + payload + "]"; }
    18      } // eIncr
    19
    20      public static class eOther extends prt.events.PEvent<Void> {
    21          public eOther() { }
    22          public Void getPayload() {
    23              throw new RuntimeException("No payload defined for event type eOther");
    24          }
    25
    26          @Override
    27          public String toString() { return "eOther"; }
    28      } // eOther
    29
    30  }
nathta@bcd0741cf59d pprojtest %
```

Note that no code is generated for the unused `eYetAnother` event.

And the remainder of the generated code, thus:

```java
nathta@bcd0741cf59d pprojtest % cat -n PGenerated/Machines.java
     1  package PGenerated;
     2
     3  /***************************************************************************
     4   * This file was auto-generated on Thursday, 23 June 2022 at 16:45:35.
     5   * Please do not edit manually!
     6   **************************************************************************/
     7
     8  import java.util.*;
     9
    10
    11  public class Machines {
    12      /* Tuples */
    13      // (a:int,b:seq[int])
    14      public static class PTuple_a_b implements prt.values.PValue<PTuple_a_b> {
...
    53      } //PTuple_a_b class definition
    54
    55
    56      public static class monitorA extends prt.Monitor {
...
    65          public monitorA() {
    66              super();
    67              addState(new prt.State.Builder(INIT_STATE)
    68                  .isInitialState(true)
    69                  .withEntry(this::Anon)
    70                  .withEvent(Events.eIncr.class, __ -> Anon_1())
    71                  .build());
    72          } // constructor
    73      } // monitorA monitor definition
    74      public static class monitorB extends prt.Monitor {
...
    83          public monitorB() {
    84              super();
    85              addState(new prt.State.Builder(INIT_STATE)
    86                  .isInitialState(true)
    87                  .withEntry(this::Anon_2)
    88                  .withEvent(Events.eOther.class, __ -> Anon_3())
    89                  .build());
    90          } // constructor
    91      } // monitorB monitor definition
    92  } // pprojtest.java class definition
nathta@bcd0741cf59d pprojtest %
```

Note that references to Event classe are now prefaced with the enclosing "namespace class", i.e. `Events.eIncr.class`.  Similarly, tuples have to refer back to their enclosing class, i.e. `Machines.PTuple_a_b`.  The natural next step would be to bring all the tuple and enum definitions out into their own file, too, but I'm going to leave that for a different PR to keep the size of these reviews small.

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/Java/MachineCodeGenerator.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/MachineCodeGenerator.cs
@@ -12,7 +12,7 @@ using Plang.Compiler.TypeChecker.Types;
 
 namespace Plang.Compiler.Backend.Java {
 
-    internal class JavaCodeGenerator : ICodeGenerator
+    internal class MachineCodeGenerator : ICodeGenerator
     {
 
         private CompilationContext _context;
@@ -22,7 +22,7 @@ namespace Plang.Compiler.Backend.Java {
         private Machine _currentMachine; // Some generated code is machine-dependent, so stash the current machine here.
 
         /// <summary>
-        /// Generates Java code for a given compilation job.
+        /// Generates Java code for a given compilation job's machine and monitor definitions.
         ///
         /// Currently, we should be able to use nested classes to put everything we need in a single
         /// Java file, in a manner similar to how the C# extractor uses namespaces.
@@ -75,16 +75,6 @@ namespace Plang.Compiler.Backend.Java {
                 foreach (var t in _globalScope.Tuples)
                 {
                     WriteNamedTupleDecl(t);
-                }
-                WriteLine();
-            }
-
-            if (_globalScope.Events.Any())
-            {
-                WriteLine("/* Events */");
-                foreach (var e in _globalScope.Events)
-                {
-                    WriteEventDecl(e);
                 }
                 WriteLine();
             }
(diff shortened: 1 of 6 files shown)


ankushdesai:
Minor comment: can we rename the `Machines` class that contains all the rest of code to the project name? (if you have other suggestions that works too). Because it contains everything not just machines. rename: `Events` to `PEvents`

Src/PCompiler/CompilerCore/Backend/Java/MachineCodeGenerator.cs
```diff
@@ -12,7 +12,7 @@
 
 namespace Plang.Compiler.Backend.Java {
 
-    internal class JavaCodeGenerator : ICodeGenerator
```
ankushdesai, line 15: Let this be JavaCodeGenerator as it contains everything else .. types, global functions, machines, monitors, UD types ...

ankushdesai (changes requested):

Src/PCompiler/CompilerCore/Backend/Java/MachineCodeGenerator.cs
dijkstracula: OK, I'll rename this to be something more specific once the patch for types is out too.

ankushdesai:
> > Minor comment: can we rename the Machines class that contains all the rest of code to the project name?
> 
> Unfortunately this becomes tricky without a lot of gnarly refactoring - basically a whole bunch of classes that were formerly decoupled would have to be tightly coupled in order to plumb that non-constant value through. I'm not convinced it's worth the engineering time (since, again, there's rarely a case for programmers to look at this file so it's mostly ergonomics that only contributors to this repo will ever see), but, very happy to take suggestions for a better name than `Machines.java` since -- at least until I pull out types into their own file, there's more in there than machines.

Lets just go with `TopDeclarations`, its bad but .... at least doesnt say its only machines.

PR #461, opened 2022-06-24
[Java] Emit proper Java enums for P enums

dijkstracula:
Initially, for simplicity, P enums were just extracted to Java as
plain old static ints, and we lost all P enum-specific type information
in Java.  Consider the very silly P program:

```
nathta@bcd0741cf59d pprojtest_enum % cat -n PSpec/foo.p
     1  enum anEnum {
     2    VALUE_ZERO,
     3    VALUE_ONE
     4  }
     5
     6  event eIncr : anEnum;
     7
     8
     9  spec monitorA observes eIncr {
    10    var myE : anEnum;
    11    start state Init {
    12      entry {
    13        myE = VALUE_ZERO;
    14      }
    15      on eIncr do (e: anEnum) {
    16        assert e == VALUE_ONE, "uh oh";
    17      }
    18    }
    19  }
nathta@bcd0741cf59d pprojtest_enum %
```

As of tag 1.13:
```java
 12     /* Enums */
 13     public static class anEnum {
 14         public static final int VALUE_ZERO = 0;
 15         public static final int VALUE_ONE = 1;
 16     }
...
 53     public static class monitorA extends prt.Monitor {
 54         private int myE = 0;
 55         public int get_myE() { return this.myE; };
...
 60         private void Anon() {
 61             myE = anEnum.VALUE_ZERO;
 62         }
 63         private void Anon_1(int e) {
 64             boolean TMP_tmp0 = false;
 65             String TMP_tmp1 = "";
 66
 67             TMP_tmp0 = e == anEnum.VALUE_ONE;
 68             TMP_tmp1 = "uh oh";
 69             tryAssert(TMP_tmp0, TMP_tmp1);
 70         }
```

This patch makes them Java enums so there's no opportunity
to use an invalid integral value where an enum is expected.

```java
// in PTypes.java
 11     /* Enums */
 12     public enum anEnum {
 13         VALUE_ZERO(0),
 14         VALUE_ONE(1);
 15         private final int value;
 16         anEnum(int i) { value = i; }
 17     }
// in PMachines.java
 11     public static class monitorA extends prt.Monitor {
 12         private PTypes.anEnum myE = PTypes.anEnum.VALUE_ZERO;
...
 18         private void Anon() {
 19             myE = PTypes.anEnum.VALUE_ZERO;
 20         }
 21         private void Anon_1(PTypes.anEnum e) {
 22             boolean TMP_tmp0 = false;
 23             String TMP_tmp1 = "";
 24
 25             TMP_tmp0 = e == PTypes.anEnum.VALUE_ONE;
 26             TMP_tmp1 = "uh oh";
 27             tryAssert(TMP_tmp0, TMP_tmp1);
 28         }
```

The downside of this is now that all enums are boxed.

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/Java/Constants.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/Constants.cs
@@ -30,7 +30,7 @@ namespace Plang.Compiler.Backend.Java
         }
 
 
-        public static readonly string MachineNamespaceName = "Machines";
+        public static readonly string MachineNamespaceName = "PMachines";
         public static readonly string MachineDefnFileName = $"{MachineNamespaceName}.java";
 
         #endregion
--- a/Src/PCompiler/CompilerCore/Backend/Java/JavaSourceGenerator.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/JavaSourceGenerator.cs
@@ -88,6 +88,7 @@ namespace Plang.Compiler.Backend.Java
 
                 /* Same with immutable types. */
                 case TypeManager.JType.JString _:
+                case TypeManager.JType.JEnum _:
                 case TypeManager.JType.JMachine _:
                     writeTermToBeCloned();
                     break;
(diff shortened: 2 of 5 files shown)


dijkstracula:
I got briefly worried about this patch and its use of `==` but since Java helpfully emits those enum values as singletons, `==` is safe to use:  [https://docs.oracle.com/javase/specs/jls/se9/html/jls-8.html#jls-8.9]("Because there is only one instance of each enum constant, it is permitted to use the == operator in place of the equals method when comparing two object references if it is known that at least one of them refers to an enum constant")

ankushdesai (approved):
This looks great! I love that we are using Java Enums now.

PR #463, opened 2022-06-27
Moved the Java values into the PSymbolic runtime

ankushdesai:


ankushdesai:
@aman-goel and @lmpick for your information.

PR #468, opened 2022-07-01
[Java] Emit FFI stub implementations

dijkstracula:
This is a partial solution to an issue I noticed this morning:

I noticed a problem with the pproj FFI example: it compiles cleanly, but would behave incorrectly should it execute: [The FFI type](https://github.com/dijkstracula/pprojexample/blob/main/PForeign/globals/Counter.java) needs to implement `deepEquals` and `deepClone`, but doesn't (it should have implemented an interface), so we would find ourselves in a situation where we emit compiling Java code that calls into runtime type checking code on the PRT side, and would do the wrong thing - essentially, the "do these values inherit from PValue" checks would fail, so identical Counters [would fall through and test as being unequal](https://github.com/p-org/P/blob/2f060bf22f1f1d449c23a99ebfa4cffc298e2267/Src/PRuntimes/PJavaRuntime/src/main/java/prt/values/Equality.java#L114). Bad!

## The best solution and why we can't do it

The best thing would be to avoid this runtime dispatching and overload `values.deepEquals()` for every possible P runtime value (for instance, `deepEquals(boolean, boolean)`, `deepEquals(int, int)`, `deepEquals(ArrayList<T>, ArrayList<T>`, and so on.). Because not every value inherits from a common interface, though, bounding the type parameters for this becomes gnarly so I think we do need some form of runtime dispatch (at least to handle `any` types).  I tried pushing this as far as I could this morning but got caught by the typechecker, so I'll keep plugging away at it over time but I'm not sure the best thing is really possible unfortunately.

## The next-best solution and why we can't mostly do it, maybe

The next best thing would have been, for the case where we didn't implement the interface correctly for two `Counter` foreign types, to throw an exception during comparison so as to say "hey, I don't know what these types are meant to be".  I think this is safe, but it changes the semantics of object comparison vis a vis the C# version, which [allows for disparate types to be compared (but evalutates to false in such cases)](https://github.com/p-org/P/blob/master/Src/PRuntimes/PCSharpRuntime/Values/PrtSet.cs#L78-L82).  So, we'd have to be sure that stricter behaviour is safe (i.e. would the P compiler frontend fail to typecheck in all cases?  I suspect not in the case of comparing two `any`s.). 

_This patch additionally includes a fix where we at least throw an exception if we can't compare two types, which changes the above described behaviour.  If we are not convinced this is safe we should remove that commit before merging._

## The next-next best thing

The _next_ best thing is at least to help future developers not make the same mistake I did: emit FFI stubs that show what the skeleton of the code _ought_ to look like.  This generates a single text file (with a .txt file extension, so Maven will not pick it up ever) demonstrating what each native FFI file should look like and where it should reside in the package hierarchy.

```
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
% 
% P <-> Java Foreign Function Interface Stubs
% 
% This file was auto-generated on Friday, 01 July 2022 at 11:22:33.  
% 
% Please separate each generated class into its own .java file (detailed throughout the file), filling
% in the body of each function definition as necessary for your project's business logic.
% 
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%


%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
% Please place and complete the following class in PForeign/globals/Counter.java :

package PForeign.globals;

public class Counter implements prt.values.PValue<Counter> {
    public Counter() { }
    
    public Counter deepClone() {
        throw new RuntimeException("Counter#deepClone() not implemented yet!"); // TODO
    }
    
    public boolean deepEquals(Counter other) {
        throw new RuntimeException("Counter#deepEquals() not implemented yet!"); // TODO
    }
    
}

%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%

% Please place and complete the following class in PForeign/myMonitor.java :

package PForeign;

import PForeign.globals.Counter;

public class myMonitor {
    public static Counter NewCounter() {
        return null; // TODO
    }
    
    public static int GetNextCount(Counter counter) {
        return 0; // TODO
    }
    
}
```

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/Java/Constants.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/Constants.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 
 namespace Plang.Compiler.Backend.Java
 {
@@ -51,8 +52,34 @@ namespace Plang.Compiler.Backend.Java
 
         #region FFI generation
 
-        internal static string FFIPackage = "PForeign";
-        internal static string GlobalFFIPackage = $"{FFIPackage}.globals";
+        private static readonly string rawFFIBanner = $@"
+P <-> Java Foreign Function Interface Stubs
+
+This file was auto-generated on {DateTime.Now.ToLongDateString()} at {DateTime.Now.ToLongTimeString()}.  
+
+Please separate each generated class into its own .java file (detailed throughout the file), filling
+in the body of each function definition as necessary for your project's business logic.
+";
+
+        internal static readonly string FFIStubFileName = "FFIStubs.txt";
+
+        internal static readonly string FFIPackage = "PForeign";
+        internal static readonly string GlobalFFIPackage = $"{FFIPackage}.globals";
+
+        // Something that is clearly not valid Java.
+        private static readonly string FFICommentToken = "%";
+
+        public static string FFICommentDivider = new StringBuilder(72).Insert(0, FFICommentToken, 72).ToString();
+
+        internal static string AsFFIComment(string line)
+        {
+            return FFICommentToken + " " + line;
+        }
+
+        internal static string[] FfiBanner => rawFFIBanner
+            .Split("\n")
+            .Select(AsFFIComment)
+            .ToArray();
 
         #endregion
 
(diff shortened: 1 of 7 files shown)


ankushdesai (approved):
The last solution you mentioned seems like the best fit. Thanks!

PR #470, opened 2022-07-05
[Java] Tuples' fields should be final

dijkstracula:
Just a nit I just noticed; after construction, fields shouldn't be modifiable.


ankushdesai (approved):
Thanks!

PR #476, opened 2022-07-12
[Java] generate exceptions from prt.exceptions package

dijkstracula:
This was a PRT refactoring change that somehow didn't get changed on the runtime side, and somehow I missed with testing.  Mea culpa.

```
nathta@bcd0741cf59d /tmp % pcl -generate:Java foo.p
Parsing ...
Type checking ...
Code generation ...
Reusing existing pom.xml
Generated PMachines.java.
Generated PEvents.java.
Generated PTypes.java.
Generated FFIStubs.txt.
Compiling foo...
nathta@bcd0741cf59d /tmp % grep 'throws prt\.' PMachines.java
        private void Anon_1() throws prt.exceptions.TransitionException {
nathta@bcd0741cf59d /tmp %
```

This patch also changes back the version number of the PRT, which I had had sitting in my git staging area without having committed it!  Whoops.  Thanks, @beyazit-yalcinkaya .

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/Java/MachineGenerator.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/MachineGenerator.cs
@@ -206,15 +206,15 @@ namespace Plang.Compiler.Backend.Java {
             List<string> throwables = new List<string>();
             if (f.CanChangeState == true)
             {
-                throwables.Add("prt.TransitionException");
+                throwables.Add("prt.exceptions.TransitionException");
             }
             if (f.CanRaiseEvent == true)
             {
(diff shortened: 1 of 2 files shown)


dijkstracula:
I think I see how I missed this: I see that [a direct import](https://github.com/p-org/P/blob/master/Src/PRuntimes/PJavaRuntime/src/test/java/tutorialmonitors/clientserver/PMachines.java#L8) was added, perhaps by IntelliJ (either automatically or out of habit on my end) when I was manually updating generated java files?

ankushdesai (approved):

PR #482, opened 2022-07-19
[Java] implement state transitions as over enums

dijkstracula:
This was an early design for the PRT that we didn't end up implementing, but I think it's worth revisiting mainly for safety reasons but also because of minor performance improvements that it offers.  

## Safety

Here is some generated Java code from the ClientServer example:

```java
190     public static class GuaranteedWithDrawProgress extends prt.Monitor {
...
195         public enum PrtStates { // Note: this is new: previously, we'd have two String literals for these two states...
196             NopendingRequests,
197             PendingReqs
198         }
199
200         public GuaranteedWithDrawProgress() {
201             super();
202             addState(prt.State.keyedOn(PrtStates.NopendingRequests)
203                 .isInitialState(true)
204                 .withEvent(PEvents.eWithDrawReq.class, p -> { Anon_3(p); gotoState(PrtStates.PendingReqs); })
205                 .build());
206             addState(prt.State.keyedOn(PrtStates.PendingReqs)
207                 .withEvent(PEvents.eWithDrawResp.class, this::Anon_4)
208                 .withEvent(PEvents.eWithDrawReq.class, p -> { Anon_5(p); gotoState(PrtStates.PendingReqs); })
209                 .build());
210         } // constructor
```

Functionally this doesn't look any different from [the currently-generated code](https://github.com/p-org/P/blob/15b50c45b39cfc8382c6b9de648675056ae374d0/Src/PRuntimes/PJavaRuntime/src/test/java/tutorialmonitors/clientserver/PMachines.java#L196-L210) - we're just not using string literals to key on states.  However, that we're using a finite type gives us an ace up our sleeves: [we can now throw an error _at `ready()` time_](https://github.com/p-org/P/blob/nathan/finite_statemap/Src/PRuntimes/PJavaRuntime/src/main/java/prt/Monitor.java#L223-L227) if there is no handler for some state: in other words, we know our state map has to be _total_.  

This use of the type system should help us gain confidence that our monitors will not crash at runtime in the middle of a run.

## Performance

For context: We have seen in microbenchmarking the PRT that there is "significant" overhead to monitors with frequent state transitions:

```
Benchmark                            (i)  Mode  Cnt     Score    Error  Units
pingpong.Benchmarks.doitFreshEvent     1  avgt   25  1707.090 ± 42.673  ns/op
pingpong.Benchmarks.doitFreshEvent    10  avgt   25   279.462 ±  1.074  ns/op
pingpong.Benchmarks.doitFreshEvent    50  avgt   25   162.016 ±  0.562  ns/op
pingpong.Benchmarks.doitFreshEvent   100  avgt   25   146.384 ±  1.099  ns/op
pingpong.Benchmarks.doitFreshEvent  1000  avgt   25   157.721 ±  3.254  ns/op
```

"Significant" goes in air quotes - yes, this is a microbenchmark and yes we can probably tolerate a microsecond of latency to switch states, but we can still do a bit of due diligence on this.

Unsurprisingly, in the same benchmark we see a nontrivial improvement in performance simply by changing the `HashMap` key type to our new Enum, avoiding a potentially-expensive string hash operation:

```
Benchmark                   (n)  Mode  Cnt     Score    Error  Units
Benchmarks.doitFreshEvent     1  avgt   25  1491.305 ± 22.903  ns/op
Benchmarks.doitFreshEvent    10  avgt   25   271.432 ±  2.376  ns/op
Benchmarks.doitFreshEvent    50  avgt   25   157.436 ±  1.679  ns/op
Benchmarks.doitFreshEvent   100  avgt   25   144.294 ±  1.187  ns/op
Benchmarks.doitFreshEvent  1000  avgt   25   163.795 ±  2.559  ns/op
````

In other words, a 12.7% improvement in the worst-case throughput.

This patch does one better: we also migrate to a `EnumMap`, which is specialised for constant-time lookup for a finite domain such as an Enum.  We see a marginal but not statistically-significant improvement over this, though:

```
Benchmark                   (n)  Mode  Cnt     Score    Error  Units
Benchmarks.doitFreshEvent     1  avgt   25  1485.592 ± 16.629  ns/op
Benchmarks.doitFreshEvent    10  avgt   25   270.649 ±  1.706  ns/op
Benchmarks.doitFreshEvent    50  avgt   25   160.748 ±  2.770  ns/op
Benchmarks.doitFreshEvent   100  avgt   25   144.863 ±  1.174  ns/op
Benchmarks.doitFreshEvent  1000  avgt   25   162.352 ±  2.080  ns/op
```

In other words, a 13.0% improvement over baseline for the worst-case test.

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/Java/Constants.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/Constants.cs
@@ -34,6 +34,8 @@ namespace Plang.Compiler.Backend.Java
         public static readonly string MachineNamespaceName = "PMachines";
         public static readonly string MachineDefnFileName = $"{MachineNamespaceName}.java";
 
+        public static readonly string StateEnumName = "PrtStates";
+
         #endregion
 
         #region Event source generation
--- a/Src/PCompiler/CompilerCore/Backend/Java/MachineGenerator.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/MachineGenerator.cs
@@ -247,7 +247,7 @@ namespace Plang.Compiler.Backend.Java {
 
         private void WriteStateBuilderDecl(State s)
         {
-            WriteLine($"addState(new prt.State.Builder<>({Names.IdentForState(s)})");
+            WriteLine($"addState(prt.State.keyedOn({Names.IdentForState(s)})");
             WriteLine($".isInitialState({TypeManager.JType.JBool.ToJavaLiteral(s.IsStart)})");
 
             if (s.Entry != null)
--- a/Src/PCompiler/CompilerCore/Backend/Java/MachineGenerator.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/MachineGenerator.cs
@@ -95,11 +95,12 @@ namespace Plang.Compiler.Backend.Java {
             WriteLine();
 
             // state identifiers
-            foreach (var s in _currentMachine.States)
+            WriteLine($"public enum {Constants.StateEnumName} {{");
+            foreach (var (state, sep) in _currentMachine.States.WithPostfixSep(","))
             {
-                //TODO: I think it's fine to use unqualified names here.  But, confirm.
-                WriteLine($"public String {Names.IdentForState(s)} = \"{s.Name}\";");
+                WriteLine($"{state.Name}{sep}");
             }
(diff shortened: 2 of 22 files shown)


ankushdesai (approved):
This is such an amazing improvement. I also love the new interface!

PR #483, opened 2022-07-20
[Java] P ints should be extracted as Java Longs

dijkstracula:
Other backends use the widest integral types available on the system.


ankushdesai (approved):

dijkstracula:
This isn't quite ready for prime time yet.  The reason is that Java will not silently downcast a long to an int, which means we need to insert some special logic for array accesses:

```java
jshell> ArrayList<Integer> a = new ArrayList<Integer>(List.of(1,2,3));
a ==> [1, 2, 3]

jshell> a.get(1L)
|  Error:
|  incompatible types: possible lossy conversion from long to int
|  a.get(1L)
|        ^^

jshell>
```

dijkstracula:
The fix for the above issue was actually trivial, so this is good for re-review.

PR #485, opened 2022-07-22
[Java] generate code for monitor providers

dijkstracula:
Previously these were written by hand, but there's no need for that.  The provider is now an inner class of the monitor, so the class name in the JAR would be something like `PMachines$GuaranteedWithDrawProgress$Supplier`.


dijkstracula:
@beyazit-yalcinkaya since your stuff needs monitor providers, any thoughts on this one?

ankushdesai (approved):

PR #487, opened 2022-07-28
Java FFI: Functions to consume the current Machine

dijkstracula:
Noticed that this was a missing part of the FFI for Java.

For functions defined within the scope of a Machine, the FFI generator now emits a call typed to [that exact Monitor type](https://github.com/dijkstracula/pprojexample/blob/nathan/fixes_example/PForeign/Monitor.java#L9).  

For top-level function definitions, we don't know for sure which Monitor is calling the function, [so we unfortunately can't be quite as precise at the type level](https://github.com/dijkstracula/pprojexample/blob/nathan/fixes_example/PForeign/P_TopScope.java#L9), so we just emit the base Monitor type instead.

This patch also makes certain state transition and assert methods that were previously protected within the Monitor type public, [so they can be called on the monitor](https://github.com/dijkstracula/pprojexample/blob/nathan/fixes_example/PForeign/Monitor.java#L12).  See [this commit message](https://github.com/p-org/P/pull/487/commits/8443e2231d2f90106f5a932f39ca6179af0d0f6e) for more details.

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/Java/Constants.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/Constants.cs
@@ -30,6 +30,7 @@ namespace Plang.Compiler.Backend.Java
             return classes.Select(pkg => $"import {pkg};");
         }
 
+        public static readonly string PGeneratedNamespaceName = "PGenerated";
 
         public static readonly string MachineNamespaceName = "PMachines";
         public static readonly string MachineDefnFileName = $"{MachineNamespaceName}.java";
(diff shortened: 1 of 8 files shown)


ankushdesai:
Oops, I might have merged PR in wrong order leading to conflicts! Please feel to merge after resolving the conflicts.

ankushdesai (approved):

dijkstracula:
Thanks, Ankush - after the rebase and force-push I need another +1 .

ankushdesai (approved):

PR #488, opened 2022-08-01
Correctly emit downcasts for covariant datatypes

dijkstracula:
Per [this TODO comment block](https://github.com/p-org/P/blob/1b69964fb0d64ca7acaf6f8d5be190703cb86940/Src/PCompiler/CompilerCore/Backend/Java/MachineGenerator.cs#L867-L875): Collections in P are covariant.  This means that because `int` is a subclass of `any`, `seq[int]` is a subclass of `seq[any]`.  We were not emitting correct Java code for cases where a downcast is expected during an assignment statement.  This patch fixes it along with some minor adjustments to extracted Java casting behaviour.

## Background

(If you are familiar with the problem with covariant collections you can skip this part)

In languages with mutable references, covariant collections are problematic:

```
jshell> String ss[] = new String[] { "Java", "Arrays", "Are", "Covariant" }
ss ==> String[4] { "Java", "Arrays", "Are", "Covariant" }

jshell> Object os[] = ss;
os ==> String[4] { "Java", "Arrays", "Are", "Covariant" }
```

This violates type safety if we decided to overwrite values in `os` (and hence in `ss`).  As far as the type system knows any Object can be stored in `os`, but of course this isn't the case, because `os == ss`.

```
jshell> os == ss;
$10 ==> true

jshell> os[1] = Long.valueOf(42)
|  Exception java.lang.ArrayStoreException: java.lang.Long
|        at (#11:1)

jshell>
```

## The bug

However, in P-land we are safer because the assignment to `os` would be from a cloned collection: the following P code will get compiled down to the following Java code:

```
  9 spec monitorA observes eIncr {
 10   var os: seq[any];
 11   start state Init {
 12     entry {
 13       var is: seq[int];
 14       os = is;
 15     }
 16   }
 17 }
```

```java
 10 public class PMachines {
 11     public static class monitorA extends prt.Monitor<monitorA.PrtStates> {
...
 21         private ArrayList<Object> os = new ArrayList<Object>();
...
 41         private void Anon() {
 42             ArrayList<Long> is = new ArrayList<Long>();
 43
 44             os = (ArrayList<Long>)prt.values.Clone.deepClone(is);
 45         }
```

Unfortunately, this isn't precisely the right code that we want to have genearted:  The Java language designers realized that because of the above issue with covariant arrays, the Collections API would be _invariant_ in its type.  Therefore, even though `j.l.Long` extends `j.l.Object`, there's no inheritance relationship between `ArrayList<Long>` and `ArrayList<Object>`.  Therefore, the above code actually fails to compile.

```
[ERROR] /private/tmp/PMachines.java:[44,18] incompatible types: java.util.ArrayList<java.lang.Long> cannot be converted to java.util.ArrayList<java.lang.Object>
```

We have "good news", though - because we lose type information at runtime, [deepClone() has no choice but to clone an `ArrayList<Long>` as an `ArrayList<Object>`](https://github.com/p-org/P/blob/master/Src/PRuntimes/PJavaRuntime/src/main/java/prt/values/Clone.java#L39-L46)!  So, the problem with the above Java code is only that we are casting it to the rval type of the assignment as opposed to the lval.

```
jshell> ArrayList<Long> is = new ArrayList<>(List.of(1L, 2L, 3L))
is ==> [1, 2, 3]

jshell> ArrayList<Object> os = (ArrayList<Object>)prt.values.Clone.deepClone(is); // deepClone returns an Object
|  Warning:
|  unchecked cast
|    required: java.util.ArrayList<java.lang.Object>
|    found:    java.lang.Object
|  ArrayList<Object> os = (ArrayList<Object>)prt.values.Clone.deepClone(is);
|                                            ^----------------------------^
os ==> [1, 2, 3]

jshell> os.add("No runtime exception here, as we are backed by an Object[] !")
$14 ==> true

jshell> os
os ==> [1, 2, 3, No runtime exception here, as we are backed by an Object[] !]

jshell>
```

The problem is that the rval `CloneExpr` has no idea that it's being used in a larger context, so there isn't a "local" way to to see that line 44 of the example program should be emitted as 

```
 44             os = (ArrayList<Object>)prt.values.Clone.deepClone(is); // Only ArrayList<Object> because that's the LHS type
```

## Solution

This issue gave me the change to rethink some parts of how deepClone() works, and I think we're in a better place in general.

### A better deepClone code emitter

The current implementation of `prt.values.Clone.deepClone` consumes and returns a plain old `Object` in all cases.  This is not necessarily incorrect, but, this means _all returned values need to be upcast to their correct type_ even if the Java type system could have known what type it is supposed to be.  You can see this in [the current generated code](https://github.com/p-org/P/blob/1b69964fb0d64ca7acaf6f8d5be190703cb86940/Src/PRuntimes/PJavaRuntime/src/test/java/testmonitors/clientserver/PMachines.java#L55) - we basically "forget" the type when we pass it to `deepClone`, and then coerce the returned type back to what we already knew it to be.

Changes to the PRT have been made such that `deepClone` now consumes an arbitrary type `T` and returns that same `T`.  In this way, we don't need to "forget and then upcast" in this way.  This lets us better leverage the Java-level typechecking to ensure we're not spuriously cloning and casting values to wildly disparate types.  And unlike before, where we just emit casts in an ad-hoc manner, here we're actually producing new AST nodes to take advantage of existing code generation mechanisms.

Note that because of type erasure, the return type of, say, the [ArrayList clone method](https://github.com/p-org/P/blob/nathan/covariant_schmovariant/Src/PRuntimes/PJavaRuntime/src/main/java/prt/values/Clone.java#L38-L47) ends up _still_ being an `ArrayList<Object>`.

Overall, this makes the Java generated code more readable - it's less noisy because the unnecessary casts are gone, and the reader will now know that when a cast is present it's because it _has_ to be there, not because the code generator was naive or silly.

Which brings us to...

### The solution to the "clone into an any" problem:

The core of this patch fixes the immediate problem described above: Line 44 is now cast to the value of `os` during assignment.

```
...
 21         private ArrayList<Object> os = new ArrayList<Object>();
...
 42             ArrayList<Long> is = new ArrayList<Long>();
 43
 44             os = ((ArrayList<Object>)((Object)prt.values.Clone.deepClone(is)));
```

Eagle-eyed reviewers will notice that this now has _two_ casts: one that downcasts the return value of deepClone to an Object, and then _upcasts_ it to an ArrayList<Object>.  This is a Java type system implementation detail: Java won't let us cast an `ArrayList<T>` to an `ArrayList<U>` for some `T extends U` for the reasons described above, so we have to be wildly unsafe (or at least rely on P's typechecker!) in the subtyping case by downcasting entirely to throw away the type parameter.  Alas.  

But, on the balance, I think if you look at the generated ClientServer code, in addition to being safer in the presence of `any`s the generated code is improved overall.

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/Java/JavaSourceGenerator.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/JavaSourceGenerator.cs
@@ -94,20 +94,21 @@ namespace Plang.Compiler.Backend.Java
                     writeTermToBeCloned();
                     break;
 
-                /* Non-boxable reference types must be cloned explicitly and then
-                 * cast to their expected type (since clone() is Object-producing). */
+                /* Collections and `any` types must be explicitly cloned with the Java P runtime's `deepClone` method;
+                 * there is one override for each of these. */
                 case TypeManager.JType.JAny _:
                 case TypeManager.JType.JMap _:
                 case TypeManager.JType.JList _:
                 case TypeManager.JType.JSet _:
-                case TypeManager.JType.JForeign _: //TODO: is this right?
-                    Write($"({t.TypeName})");
                     Write($"{Constants.PrtDeepCloneMethodName}(");
                     writeTermToBeCloned();
                     Write(")");
                     break;
 
-                /* JNamedTuples have a copy constructor. */
+                /* JNamedTuples and foreign types extend prt.values.PValue, and thus have an explicit `.deepEquals()`
+                 * method.  (We could have just as easily passed these to the runtime's `deepClone` method, but this
+                 * saves us a type dispatch). */
+                case TypeManager.JType.JForeign _:
                 case TypeManager.JType.JNamedTuple _:
                     writeTermToBeCloned();
                     Write(".deepClone()");
@@ -128,7 +129,7 @@ namespace Plang.Compiler.Backend.Java
         }
     }
 
-    static class IEnumerableExtensions
+    static class EnumerableExtensions
     {
 
         /// <summary>
--- a/Src/PCompiler/CompilerCore/Backend/Java/MachineGenerator.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/MachineGenerator.cs
@@ -562,7 +562,6 @@ namespace Plang.Compiler.Backend.Java {
                     break;
 
                 case SeqAccessExpr seqAccessExpr:
-                    // TODO: do we need to think about handling out of bounds exceptions?
                     WriteExpr(seqAccessExpr.SeqExpr);
                     Write($".{t.MutatorMethodName}(");
                     WriteExpr(seqAccessExpr.IndexExpr);
--- a/Src/PCompiler/CompilerCore/Backend/Java/MachineGenerator.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/MachineGenerator.cs
@@ -527,6 +527,20 @@ namespace Plang.Compiler.Backend.Java {
 
             IPExpr rval = assignStmt.Value;
 
+            // In the case where the types of each side of the assignment differ due to subtyping, we need to ensure
+            // that we cast the given rval to the lval's type.  For example, if we are assigning a `seq[int]` to a
+            // `seq[any]`, then we need to first downcast the `ArrayList<Long>` to an Object, and then _re-upcast_ it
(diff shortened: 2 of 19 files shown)


ankushdesai (approved):
I love, all the efforts you have put in generating the correct/succint code

PR #493, opened 2022-08-10
[Java] nit: invert inverted conditionals

dijkstracula:
This is not a correctness or performance issue but I've been staring at some awkward generated code today and this has been driving me nuts!

Here are two event handlers:

```
  7     on e1 do {
  8       var i: int;
  9       i = 0;
 10       while (i < sizeof(numbers)) {
 11         i = i + 1;
 12       }
 13     }
 14     on e2 do {
 15       if (true) {} else {}
 16     }
```

Currently, the loop in `e1` and "useless if" in e2 is emitted in a correct but awkward way:

```java
 49             while ((true)) {
 50                 TMP_tmp0 = numbers.size();
 51                 TMP_tmp1 = i < TMP_tmp0;
 52                 TMP_tmp2 = TMP_tmp1;
 53                 if (TMP_tmp2) {} else
 54                 {
 55                     break;
 56                 }
 57                 TMP_tmp3 = i + 1L;
 58                 i = TMP_tmp3;
 59             }
...
 62         private void Anon_1() {
 63             if ((true)) {}}
```

For the first case, we rearrange the conditional to its negation if the "then" body is empty but the "else" is not.  (The useless parens around the boolean literal are also removed.)

```java
 49             while (true) {
 50                 TMP_tmp0 = numbers.size();
 51                 TMP_tmp1 = i < TMP_tmp0;
 52                 TMP_tmp2 = TMP_tmp1;
 53                 if (!(TMP_tmp2)) {
 54                     break;
 55                 }
 56                 TMP_tmp3 = i + 1L;
 57                 i = TMP_tmp3;
 58             }
```

For the latter case, where both branches are empty, we can elide the statement altogether.

```java
 61         private void Anon_1() {
 62         }
```

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/Java/MachineGenerator.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/MachineGenerator.cs
@@ -408,26 +408,38 @@ namespace Plang.Compiler.Backend.Java {
                     break;
 
                 case IfStmt ifStmt:
+                    if (!ifStmt.ThenBranch.Statements.Any() && !ifStmt.ElseBranch.Statements.Any())
+                    {
+                        break;
+                    }
+
+                    if (!ifStmt.ThenBranch.Statements.Any() && ifStmt.ElseBranch.Statements.Any())
+                    {
+                        UnaryOpExpr inverted = new UnaryOpExpr(
+                            ifStmt.SourceLocation,
+                            UnaryOpType.Not,
+                            ifStmt.Condition
+                        );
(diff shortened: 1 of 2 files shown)


dijkstracula:
Actually, I wonder whether eliding `if (true) {} else {}` is the correct thing - the conditional expression could be, like, a function call with side-effects?

dijkstracula:
Actually, the more I think about this I think this is the wrong place to do this - as it stands, negating the temp like that means we kind of break SSA?

PR #498, opened 2022-08-17
[PUnit] do not build as part of build.sh

dijkstracula:
The previous patch configured the build script to also build PUnit.  However, I realised this morning that this implicitly required that the PRT is installed in the user's local Maven repository; from a clean ~/.m2 directory, I get:

```
$ ./build.sh
...
[ERROR] Failed to execute goal on project punit: Could not resolve dependencies for project org.example:prts
andbox:jar:1.0-SNAPSHOT: Could not find artifact p.runtime:PJavaRuntime:jar:1.0-SNAPSHOT -> [Help 1]
```

One simple change would be to, at build time, install that dependency - we're building it anyway, so we just need to do a `mvn install` rather than `mvn compile`.  However, I worry about implicitly installing dependencies in a build script: in particular, if a user is actively working on the PRT and also re-building P, I'm not exactly sure which version will get stomped over unless the version is bumped explicitly.

So, given that PUnit is not a critical part of the P project, just don't build it.  If it becomes important later, it's easy to fix.

### DIFF
--- a/Bld/build.sh
+++ b/Bld/build.sh
@@ -11,10 +11,7 @@ pushd ..
 git submodule update --init --recursive
 
 echo -e "${ORANGE} ---- Building the Java P runtime ----${NOCOLOR}"
-mvn clean compile -f ./Src/PRuntimes/PJavaRuntime/pom.xml
-
-echo -e "${ORANGE} ---- Building PUnit ----${NOCOLOR}"
-mvn clean compile -f ./Tst/PUnit/pom.xml
+mvn compile -q -f ./Src/PRuntimes/PJavaRuntime/pom.xml
 
 echo -e "${ORANGE} ---- Building the PCompiler ----${NOCOLOR}"
 # Run the build!
--- a/Bld/build.sh
+++ b/Bld/build.sh
@@ -5,8 +5,11 @@ GREEN='\033[0;32m'
 ORANGE='\033[0;33m'
 set -e
 
+BLD_PATH=$(dirname $0)
+pushd $BLD_PATH/..
+
 echo -e "${ORANGE} ---- Fetching git submodules ----${NOCOLOR}"
-pushd ..
+
 # Initialize submodules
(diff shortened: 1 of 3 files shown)


dijkstracula:
A second small change that has driven me nuts all summer:

Currently, the build script doesn't work unless it's run from `Bld/`.

```
$ ./Bld/build.sh
 ---- Fetching git submodules ----
~/code ~/code/P
fatal: not a git repository (or any of the parent directories): .git
$
```

Now it changes into the script's parent directory, which is independent of where the script is run from.

```
$ # run it from the Bld/ directory
$ cd Bld
$ ./build.sh
~/code/P ~/code/P/Bld
 ---- Fetching git submodules ----
 ---- Building the Java P runtime ----
^C

$ # or the root of the P repo
$ cd ..
$ ./Bld/build.sh
~/code/P ~/code/P
 ---- Fetching git submodules ----
 ---- Building the Java P runtime ----
^C

$ # Or anywhere
$ cd /tmp
$ ~/code/P/Bld/build.sh
~/code/P /tmp
 ---- Fetching git submodules ----
 ---- Building the Java P runtime ----
^C
$
```

Bld/build.sh
```diff
@@ -5,16 +5,16 @@ GREEN='\033[0;32m'
 ORANGE='\033[0;33m'
 set -e
 
+BLD_PATH=$(dirname $0)
+pushd $BLD_PATH/..
+
 echo -e "${ORANGE} ---- Fetching git submodules ----${NOCOLOR}"
-pushd ..
+
 # Initialize submodules
```
ankushdesai, line 13: Thanks a lot, I was lazy to never fix this.

ankushdesai (approved):

Bld/build.sh
dijkstracula: Yeah, I just ended up getting into the habit of calling `dotnet` directly instead :)

PR #502, opened 2022-09-14
[PSym] Updates loggers to log4j2

aman-goel:
This PR is for updating PSym loggers to log4j2.


ankushdesai (approved):
Thanks for fixing these!

PR #531, opened 2022-11-29
Update FFI file to remove machien as first param

soumyasmruti:
Removed the monitorType machine line the writes machine as the first parameter in FFIStubGenerator.cs.


ankushdesai (approved):
Thanks for this change!

PR #559, opened 2023-02-10
[CLI] Updates commandline options

aman-goel:
Updates include:
- Updates CLI options to remove `--target` and `--generate` options. Instead, `--target` is renamed to `--projname` and `--mode` is used to configure the compiler/checker backend (default: `bugfinding`). Currently available modes include: `bugfinding`, `pobserve`.
- Adds support for searching for `*.dll` in the current directory with `p check` command when path is not given
- Adds fallback support for searching for and adding local P files in the current directory with `p compile` command when no P file is given through `*.pproj` and `--pfiles` options.
- Improves error messages in commandline configuration

### DIFF
--- a/Src/PCompiler/PCommandLine/Options/PCheckerOptions.cs
+++ b/Src/PCompiler/PCommandLine/Options/PCheckerOptions.cs
@@ -2,9 +2,10 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using PChecker;
-using PChecker.IO;
 using PChecker.IO.Debugging;
 using Plang.Parser;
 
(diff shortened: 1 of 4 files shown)


Src/PCompiler/PCommandLine/Options/PCheckerOptions.cs
```diff
@@ -238,5 +248,41 @@ private static void SanitizeConfiguration(CheckerConfiguration checkerConfigurat
...
+                    "cover" => "*-jar-with-dependencies.jar",
+                    "pobserve" => "*-jar-with-dependencies.jar",
+                    _ => "*.dll"
+                };
+                
+                var files = Directory.GetFiles(Directory.GetCurrentDirectory(), filePattern, SearchOption.AllDirectories);
+
+                foreach (var fileName in files)
+                {
+                    if (fileName.EndsWith("PCheckerCore.dll") || fileName.EndsWith("PCSharpRuntime.dll")) continue;
```
ankushdesai, line 272: How about searching for a file with the project name? Which is what we use as the generated dll name?

ankushdesai (changes requested):
Please drop the unused field error as well.

Src/PCompiler/PCommandLine/Options/PCheckerOptions.cs
aman-goel: `p check` command does not know the project name as it doesn't have the option `--projname` right now.
Do you suggest adding such an option?

aman-goel: Can you clarify what you mean by "unused field error"?

PR #571, opened 2023-03-29
[P 2.0] Updates

aman-goel:
Hi @ankushdesai,

I added the following updates to improve P 2.0 release.
- Disables `--mode` option in release mode (only allowed when `--debug` is passed for now)
- Update portfolio checker script to P 2.0 CLI
- Minor: Updates folder hierarchy of generated targets in PCompiler
- Minor: Adds some updates to GitHub docs

Let me know of any suggested changes.

--
@aman-goel

### DIFF
--- a/Src/PChecker/CheckerCore/CheckerCore.csproj
+++ b/Src/PChecker/CheckerCore/CheckerCore.csproj
@@ -6,5 +6,6 @@
         <GenerateDocumentationFile>true</GenerateDocumentationFile>
         <OutputPath>$(PSdkFolder)\Binaries</OutputPath>
         <TargetFramework>net6.0</TargetFramework>
+        <Version>2.0.4</Version>
     </PropertyGroup>
 </Project>
--- a/Src/PChecker/CheckerCore/CheckerCore.csproj
+++ b/Src/PChecker/CheckerCore/CheckerCore.csproj
@@ -6,6 +6,5 @@
         <GenerateDocumentationFile>true</GenerateDocumentationFile>
         <OutputPath>$(PSdkFolder)\Binaries</OutputPath>
         <TargetFramework>net6.0</TargetFramework>
-        <Version>2.0.4</Version>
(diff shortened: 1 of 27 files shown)


Src/PChecker/CheckerCore/CheckerCore.csproj
```diff
@@ -6,5 +6,6 @@
         <GenerateDocumentationFile>true</GenerateDocumentationFile>
         <OutputPath>$(PSdkFolder)\Binaries</OutputPath>
         <TargetFramework>net6.0</TargetFramework>
+        <Version>2.0.4</Version>
```
ankushdesai, line 9: Why are we hard-coding it here?

Src/PCompiler/CompilerCore/CompilerConfiguration.cs
```diff
@@ -53,6 +54,7 @@ public CompilerConfiguration()
             OutputLanguage = outputLanguage;
             Backend = TargetLanguage.GetCodeGenerator(outputLanguage);
             ProjectDependencies = projectDependencies ?? new List<string>();
+            EnableDebugging = false;
```
ankushdesai, line 57: There was a misunderstanding. Instead of a command line flag. I was referring to a build macro that gets set based on the dotnet build debug or release

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives

We would like to use
```
if #DEBUG
```

 then 
```
dotnet -c Release 
or 
dotnet -c Debug
```

ankushdesai (changes requested):
Please look at my comment regarding debug flag

Src/PChecker/CheckerCore/CheckerCore.csproj
aman-goel: Corrected in f4e0178a4f0df7847656d5262104f2154ed71df8

PR #573, opened 2023-04-05
[2.0] Updates

aman-goel:
Updates include:
- Programmatically set the dependency version in the generated `<project-name>.csproj`
- Minor corrections to PSym (relating to union-find with dynamic machine creation)

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/CSharp/CSharpCodeGenerator.cs
+++ b/Src/PCompiler/CompilerCore/Backend/CSharp/CSharpCodeGenerator.cs
@@ -46,8 +46,16 @@ namespace Plang.Compiler.Backend.CSharp
                     }
                     foreignInclude += "  </ItemGroup>";
                 }
-
                 csprojTemplate = csprojTemplate.Replace("-foreign-include-", foreignInclude);
+
+                string versionString = "2.*";                
+                var version = typeof(CSharpCodeGenerator).Assembly.GetName().Version;
+                if (version is not null && version.Major > 1)
+                {
+                    versionString = version.ToString();
+                }
+                csprojTemplate = csprojTemplate.Replace("-version-", versionString);
+
                 File.WriteAllText(csprojPath, csprojTemplate);
             }
 
--- a/Src/PRuntimes/PSymRuntime/pom.xml
+++ b/Src/PRuntimes/PSymRuntime/pom.xml
@@ -428,7 +428,7 @@
     <maven.compiler.target>11</maven.compiler.target>
     <log4j2.configurationFile>${project.basedir}/src/main/resources/log4j2.xml</log4j2.configurationFile>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <revision>1.0.0</revision>
+    <revision>1.0.1</revision>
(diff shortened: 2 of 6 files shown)


Src/PCompiler/CompilerCore/Backend/CSharp/CSharpCodeGenerator.cs
```diff
@@ -46,8 +46,16 @@ public void Compile(ICompilerConfiguration job)
                     }
                     foreignInclude += "  </ItemGroup>";
                 }
-
                 csprojTemplate = csprojTemplate.Replace("-foreign-include-", foreignInclude);
+
```
ankushdesai, line 50: Nice! I like it!

Src/PRuntimes/PSymRuntime/pom.xml
```diff
@@ -428,7 +428,7 @@
     <maven.compiler.target>11</maven.compiler.target>
     <log4j2.configurationFile>${project.basedir}/src/main/resources/log4j2.xml</log4j2.configurationFile>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <revision>1.0.0</revision>
+    <revision>1.0.1</revision>
```
ankushdesai, line 431: Do we want to make this versioning consistent with the P versions? Just like we do it for all the other P nuget packages?

Src/PRuntimes/PSymRuntime/src/main/java/psym/valuesummary/util/ValueSummaryUnionFind.java
```diff
@@ -32,9 +32,30 @@ public Map<Set<PrimitiveVS>, Guard> getLastUniverseMap() {
         for (Set<PrimitiveVS> set : lastDisjointSet) {
             lastUniverseMap.put(set, universe.get(find(set.iterator().next())));
         }
+        assert(sanityCheck(lastUniverseMap));
         return lastUniverseMap;
     }
 
+    public boolean sanityCheck(Map<Set<PrimitiveVS>, Guard> universeMap) {
```
ankushdesai, line 39: Yo! is this a bug fix?

Src/PRuntimes/PSymRuntime/pom.xml
aman-goel: Noted! We will do this when we make PSym release.

Src/PRuntimes/PSymRuntime/src/main/java/psym/valuesummary/util/ValueSummaryUnionFind.java
aman-goel: Yes. When dynamic machine creation is involved, [ValueSummaryUnionFind](https://github.com/p-org/P/blob/1fda3058ca3b0bcd28efe494d3d6528bb12a3052/Src/PRuntimes/PSymRuntime/src/main/java/psym/valuesummary/util/ValueSummaryUnionFind.java#L12) breaks and was no longer producing disjoint sets of guards. I corrected this and added this sanity check.

ankushdesai (approved):

PR #598, opened 2023-06-03
PCheckerOptions fix file path search

mlh758:
Everything seems to work fine on Windows except for actually running the spec once compiled. I noticed the checker options were hard coded with `/` so I swapped it to use the character from `Path`.

I can run specs on Windows now!

Would you be open to another PR that adds a "Developing P" section to the readme? I had to look up how to install local versions of `dotnet` tools which took longer than coding up this fix. You might also assume that this is relatively common knowledge and not want to duplicate it in your repository though.


ankushdesai (commented):
Thanks! for the change.

Yes, we would definitely appreciate any help on strengthening the documentation. Looking forward to the PR!

ankushdesai (approved):

aman-goel (approved):

PR #600, opened 2023-06-05
[CI] Fix Tutorial GitHub Action

aman-goel:
GitHub action for Tutorials was picking up the latest P version on NuGet and using it to run the GitHub action.
What we instead need in our CI is that P is built from the current code commit (like all other GitHub actions).
This PR adds support for this.

Also, the shell script [check.sh](https://github.com/p-org/P/blob/master/Src/Scripts/TutorialsChecker/check.sh#L30) was incorrectly parsing the test cases (due to the recent P change "[PTool]" became "~~ [PTool]"). Corrected.

### DIFF
--- a/.github/workflows/tutorials.yml
+++ b/.github/workflows/tutorials.yml
@@ -18,11 +18,9 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 11
-    - name: Build P from source
-      run: dotnet build --configuration Release
+    - name: Install P as a tool
+      run: dotnet tool install --global p
     - name: Compile Tutorials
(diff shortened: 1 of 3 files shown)


ankushdesai:
I purposely want the tutorials to use the public released version of P. This will test our commandline interface and the public installation, etc.

ankushdesai:
There is something else that is flaky in my current check.sh;

aman-goel:
> I purposely want the tutorials to use the public released version of P. This will test our commandline interface and the public installation, etc.

Okay. Note that this means we would only get info on whether or not all tutorials are working on a new release **after** we release the new P version.

aman-goel:
> There is something else that is flaky in my current check.sh; Aah, got it. It was the ~~ change. Can you update the CR to only fix that?

Done!

ankushdesai:
Thanks!!!

ankushdesai (approved):

PR #601, opened 2023-06-07
Fix tutorials for the renamed test case

y-taka-23:
In the Client Server tutorial, the name of the test case for `AbstractServer` should not be `tcSingleClientAbstractServer` but `tcAbstractServer`, as specified in `Tutorial/1_ClientServer/PTst/Testscript.p`.


ankushdesai (approved):
Thanks for these fixes!

ankushdesai:
Pushed the changes out!

PR #609, opened 2023-06-23
[PSym] Updates

aman-goel:
Updates include:
- [PSym] Adds the first version of symmetry-aware exploration with symbolic search (needs optimizations)
- [PSym] Refactors code to separate out symbolic and explicit search
- [PSym] Cleanup obsolete stats collection for symbolic
- [PSym] Changes default for explicit search to use fast state hashing instead of string-based exact caching


ankushdesai (approved):
Good to see this moving towards a cleaner implementation!

PR #623, opened 2023-07-18
Feature/log json formatter

haoran-wen:
\+ Created JSON writer to write a JSON log output
\+ Vector clock generator built in JSON writer
\+ Added `PJsonFormatter.cs`, `ActorRuntimeLogJsonFormatter.cs`, and `JsonWriter.cs`
\+ Checker configuration outputs JSON log on default (no need to include any flag in p checker CLI)

**NOTE**: JSON parsing of the payload (if any is sent in an event) is only formatted nicely for named tuple types as of right now.


ankushdesai:
Looks great! THANKS!

ankushdesai (approved):

PR #624, opened 2023-07-19
[PCover] Add CI

aman-goel:
Updates include:

[PCover]
- Add GitHub action for explicit-state checker
- Improve error reported due to detected cycles in a schedule
- Cleanup when liveness checks are enforced

[PSym]
- Improve fix point detection (enabled via `--state-caching exact` CLI option)
- Minor correction to corner-case symbolic set comparison

### DIFF
--- a/Src/PRuntimes/PSymRuntime/src/test/java/psym/TestSymbolicRegression.java
+++ b/Src/PRuntimes/PSymRuntime/src/test/java/psym/TestSymbolicRegression.java
@@ -26,7 +26,7 @@ public class TestSymbolicRegression {
   private static final String outputDirectory = "output/testCases";
   private static final List<String> excluded = new ArrayList<>();
   private static String mode = "verification";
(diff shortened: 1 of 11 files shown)


Src/PRuntimes/PSymRuntime/src/test/java/psym/TestSymbolicRegression.java
```diff
@@ -26,7 +26,7 @@ public class TestSymbolicRegression {
   private static final String outputDirectory = "output/testCases";
   private static final List<String> excluded = new ArrayList<>();
   private static String mode = "verification";
-  private static String runArgs = "--seed 0 --timeout 60 --max-steps 300";
+  private static String runArgs = "--seed 0 --timeout 60 --iterations 100 --max-steps 300";
```
ankushdesai, line 29: why is the seed fixed?

ankushdesai (approved):

PR #635, opened 2023-08-01
[PCompiler] Fix stale version mismatch

aman-goel:
Force non-incremental flag when building C# generated code to avoid stale version mismatch from different versions of P. 

Before the fix, running `p compile` followed by `pl compile` should throw a compile error (like https://github.com/p-org/peasy-ide-vscode/issues/23) if local P version (i.e., `pl`) does not match official (i.e., `p`). Similar issue can happen when updating P and using latest P release version on project previously built with an old release.

After the fix, the error should go away, since `--no-incremental` forces the C# generated code to perform `clean and build` instead of just `build`, thus ignoring previous dotnet object files that can trigger such error due to P .dll version mismatch.


ankushdesai (approved):
Thanks! Please confirm that this fixes the issue.

ankushdesai:
Thanks! Please confirm that this fixes the issue.

PR #638, opened 2023-08-03
[PSym/PCover] Add fixpoint, symmetry search

aman-goel:
[PSym]
- Adds symbolic state caching with support for pruning explored states and fix point detection
- Updates CLI to include new symbolic strategies: symbolic-fixpoint and symbolic-iterative (choices bounded to 2 per iteration)
- Cex replayer: corrects behavior with dynamic machine creation in PSym mode

[PCover]
- Adds core algorithm to track dependent fields and perform symmetry-aware search with them

[CI]
- Minor updates relating to configuring checker run for PSym

### DIFF
--- a/.github/workflows/psymit.yml
+++ b/.github/workflows/psymit.yml
@@ -37,4 +37,4 @@ jobs:
       run: ./scripts/build.sh
     - name: Test PSym
       working-directory: Src/PRuntimes/PSymRuntime
-      run: mvn test -Dpsym.args="-sb 2 -db 2"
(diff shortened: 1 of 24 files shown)


ankushdesai (approved):
I would like to see a better way of passing commandline arguments to psym. something like if `--mode verification` then all the arguments except for general ones are automatically sent to the psym tool.

aman-goel:
> I would like to see a better way of passing commandline arguments to psym. something like if `--mode verification` then all the arguments except for general ones are automatically sent to the psym tool.

Makes sense. Will update as a separate PR.

PR #639, opened 2023-08-07
Added JSON representation of payload of all types, beyond named tuples

haoran-wen:
\+ Added JSON representation of payload all P data types, beyond just named tuples

Added `object ToDict()` method for all `IPrtValues`

JSON payload should you represent all valid P data types, including nested data types.

#### `tuple`
```
{
   [index]: value,
   ...
}

Example: (0, false, "foo") becomes
{
   0: 0,
   1: false,
   2: "foo",
}
```

####  `named tuple`
```
{
   [name]: value,
   ...
}

Example: (client=Client(7), trans=(key=8, val=8, transId=12)) becomes
{
   "client": "Client(7)",
   "trans": {
      "key": 8,
      "val": 8,
      "transId": 12
   }
}
```

#### `int`, `bool`, `float`, `string`
```
[value itself]

Example: int 8 becomes
8
bool false becomes
false
etc...
```

#### `enum`
```
{
   [name]: [int value],
   ...
}

Example: { ERROR = 500, SUCCESS = 200, TIMEOUT = 400; } becomes
{
   "ERROR": 500,
   "SUCCESS": 200,
   "TIMEOUT": 400,
}
```

#### `map`
```
{
   [key]: [value],
   ...
}

Example: (<key1 -> 2>, <key2 -> 4>) becomes
{
   "key1": 2,
   "key2": 4,
}
```

#### `seq`, `set`
```
[ [value], ... ]

Example: Both (<5>, <"value">, <false>) (set) and (5, "value", false) (seq) becomes
[ 5, "value", false ]
```

### DIFF
--- a/Src/PChecker/CheckerCore/Actors/Logging/JsonWriter.cs
+++ b/Src/PChecker/CheckerCore/Actors/Logging/JsonWriter.cs
@@ -129,53 +129,64 @@ namespace PChecker.Actors.Logging
 
         /// <summary>
         /// Converts payload from json to string representation.
+        ///
+        /// Recursively builds string in case of nested data types.
         /// </summary>
-        /// <param name="eventPayload">Of type Dictionary&lt;string, string&gt;: JSON representation of payload.</param>
+        /// <param name="eventPayload">Of type object: JSON representation of payload.</param>
         /// <returns>string: string representation of payload.</returns>
         private static string? ConvertPayloadToString(object? eventPayload)
         {
-            // If no payload, return empty string.
-            if (eventPayload is null)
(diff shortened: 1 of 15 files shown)


soumyasmruti:
Great work, how would foreign types or custom data types will look ? Can you add some examples ?

haoran-wen:
> Great work, how would foreign types or custom data types will look ? Can you add some examples ?

Since `foreign` and `any` is anything that user defines out of the scope of the types explicitly defined by P, the JSON representation is just the string representation of them as of now.

aman-goel (approved):
Looks good!

PR #648, opened 2023-08-31
Vector clock updated for OnMonitorProcessEvents

haoran-wen:
#### Update
For `OnMonitorProcessEvents`, the monitor's vector clock will be connected to the `senderMachine`'s vector clock so there is some timeline association between when the monitor processed the event and when the senderMachine sent the event. 

#### Previous Issue
The issue before was that since there was no association between the monitor's and the senderMachine's events, all the monitor's individual events get's stacked.

Screenshot 2023-08-30 at 5 16 46 PM

#### Solution
Now, the individual events are expanded to better align with when the event was actually sent.

Screenshot 2023-08-30 at 5 15 01 PM

Screenshot 2023-08-30 at 5 14 25 PM


ankushdesai (approved):
Thanks a lot for this awesome work!!!

PR #654, opened 2023-09-14
Fix p check replay bug and stately code generator bug

mchadalavada:
This PR fixes the below issues:
1. The P output files were moved before replay command is run, causing a "File not found" error
2. The stately generator code was generating duplicate handlers for a state


ankushdesai (approved):
Looks great!

ankushdesai (changes requested):
Fix the error, make sure things build and pass test locally.

ankushdesai (approved):

PR #655, opened 2023-09-18
Update docs and fix stately generator code bug

mchadalavada:
1. Update "IDE for P" section in the P documentation
2. Fix stately code generator bug that throws key not found error


ankushdesai (approved):
Thanks! Looks great.

PR #659, opened 2023-09-22
Add logic to remove null key-value pairs in trace json

mchadalavada:
The Json serializer removes keys with null values in the parent attributes only. If the one of the attributes has children and if the child has a key with null value, that child key is not removed. This resulted in the trace json having null values and the visualizer failed to show the trace.

The current fix recursively traverses through the event payload and removes the keys that have null values


ankushdesai (commented):
Why are we removing keys with `null` values instead of `replacing them with a string null`

ankushdesai (approved):

PR #662, opened 2023-10-04
[C#] Correct error in coverage reporting

aman-goel:


ankushdesai (changes requested):
Does this also fix the issue of in the coverage report, we dont print `PImplementation.X`

aman-goel:
> Does this also fix the issue of in the coverage report, we dont print `PImplementation.X`

Done via [aef7289](https://github.com/p-org/P/pull/662/commits/aef7289387c2358720f498be05da8c7afcb7e0b2)

Src/PChecker/CheckerCore/Coverage/ActivityCoverageReporter.cs
```diff
@@ -349,5 +353,15 @@ private static string GetMachineId(string nodeId)
 
         private static string GetStateId(string machineName, string stateName) =>
             string.Format("{0}::{1}", stateName, machineName);
+        
+        private static string GetSanitizedName(string name)
```
ankushdesai, line 357: Instead of doing this can you do `name.split(".").Last()`

aman-goel: Done via [d752c52](https://github.com/p-org/P/pull/662/commits/d752c5233a417fd23f38ac20bb8b074c907881e7)

PR #500, opened 2022-09-12
Upgrades to PSym

aman-goel:
Hi,

I have added several upgrades to PSym. Here is a summary of major changes:

- Adds state-space coverage reporting
- Adds foreach to while transformer
- Add support for specifying test method
- Support map value summary with tuple keys
- Adds initial support for allowing recursive functions
- Adds MonoSAT backend
- Adds memory/time monitors
- Updates loggers, scripts, and readme
- Updates storing state in file and resuming

Ankush (@ankushdesai), it would be great if you can review and let me know of any suggested changes.

--
Aman
@aman-goel

PR #684, opened 2023-11-26
[PCompiler] Correct foreach -> while transformation 

aman-goel:
[PCompiler]
Corrects foreach to while transformation to account for possible continue statement in loop body

[Tst]
Adds regression test for foreach with continue

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/CSharp/CSharpCodeGenerator.cs
+++ b/Src/PCompiler/CompilerCore/Backend/CSharp/CSharpCodeGenerator.cs
@@ -1435,7 +1435,7 @@ namespace Plang.Compiler.Backend.CSharp
                     break;
 
                 case IntLiteralExpr intLiteralExpr:
-                    context.Write(output, $"((PrtInt){intLiteralExpr.Value})");
+                    context.Write(output, $"((PrtInt)({intLiteralExpr.Value}))");
                     break;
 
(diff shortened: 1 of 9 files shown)


ankushdesai (changes requested):
At some point the code generation in the C# was changed from `foreach` statement to this `while` loop logic. I would like to know why?

P foreach can be simply converted to C# `foreach` can we roll back to that change?

Also, can you add more regressions both positive and negative.
For break, continue and combination of break and continue used at the same time.

aman-goel:
> At some point the code generation in the C# was changed from `foreach` statement to this `while` loop logic. I would like to know why?
> 
> P foreach can be simply converted to C# `foreach` can we roll back to that change?
> 
> Also, can you add more regressions both positive and negative. For break, continue and combination of break and continue used at the same time.

We added foreach to while IR transformation in #500 to support foreach in PSym/PCover. Originally, it was done only for PSym backend as a language/backend specific transformation. In the PR #500, based on the suggested comments, we enabled this transformation to all backends (including for PObserve) via 1b33a4628d38b61cc2deb68ec6314925fa72855b.

There was a minor bug in this transformation when the loop body has a continue statement (which this PR corrects).

Also added more +/- regression tests for foreach with break/continue via 82d768e681c27b6bde484d9cc2c78e05f881501f.

ankushdesai (approved):

PR #685, opened 2023-11-30
Add option to enable/disable debug logs in pobserve mode

mchadalavada:
This PR includes the following changes:
1. Converts P spec print statement to java logger statements instead of System.out when run in pobserve mode
2. Provides an option to enable or disable these log statements through command line option (--debug)

Usage: p compile --mode pobserve --debug (true|false)

### DIFF
--- a/Src/PChecker/CheckerCore/SystematicTesting/TestingEngine.cs
+++ b/Src/PChecker/CheckerCore/SystematicTesting/TestingEngine.cs
@@ -592,6 +592,36 @@ namespace PChecker.SystematicTesting
         }
 
         /// <summary>
+        /// Returns an object where the keys with null values are removed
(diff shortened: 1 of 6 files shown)


aman-goel:
Hi @mchadalavada,

Changes look good to me.

I do have a minor suggestion: Can we make the `--debug` option of boolean type without any arguments, i.e., `Usage: p compile --mode pobserve --debug`? This is how the `p check` command also implements the `--debug` flag (check [here](https://github.com/p-org/P/blob/master/Src/PCompiler/PCommandLine/Options/PCheckerOptions.cs#L45)). Also, let's make this flag hidden from P command-line help, since not meant for general-purpose use.

aman-goel (approved):

PR #686, opened 2023-12-30
Remove invalid '>' from while condition `(i < sizeof(sq)>)` in `P Expressions` doc

komamitsu:
Found a typo in https://p-org.github.io/P/manual/expressions/#sizeof.

```
while (i < sizeof(sq)>) {
```

This PR fixes it.


aman-goel (approved):
Thanks for the correction 👍

PR #702, opened 2024-03-07
[PCompiler] Remove test cases from imported projects

aman-goel:


Src/PCompiler/CompilerCore/TypeChecker/Scope.cs
```diff
@@ -32,9 +34,11 @@ public class Scope
         private readonly IDictionary<string, TypeDef> typedefs = new Dictionary<string, TypeDef>();
         private readonly IDictionary<string, Variable> variables = new Dictionary<string, Variable>();
 
-        private Scope(ITranslationErrorHandler handler, Scope parent = null)
+        private Scope(ITranslationErrorHandler handler, ILocationResolver locationResolver, IList<string> projectDependencies, Scope parent)
```
ankushdesai, line 37: make `parent = null` so that its not a required argument and then drop the `null`

Src/PCompiler/CompilerCore/TypeChecker/Analyzer.cs
```diff
@@ -161,7 +161,7 @@ private static void ApplyPropagations<T>(IEnumerable<Function> functions, params
 
         private static Scope BuildGlobalScope(ICompilerConfiguration config, PParser.ProgramContext[] programUnits)
         {
-            var globalScope = Scope.CreateGlobalScope(config.Handler);
+            var globalScope = Scope.CreateGlobalScope(config.Handler, config.LocationResolver, config.ProjectDependencies);
```
ankushdesai, line 164: why dont we pass the config itself instead of passing 3 different parameters?

ankushdesai (changes requested):

Src/PCompiler/CompilerCore/TypeChecker/Scope.cs
aman-goel: Done via 6758a4af0c3e39ef87622405bb8c78c12dea7a76

Src/PCompiler/CompilerCore/TypeChecker/Analyzer.cs
aman-goel: Good point!
Done via 6758a4af0c3e39ef87622405bb8c78c12dea7a76.

ankushdesai (approved):

PR #708, opened 2024-03-26
[PExplicit] Initial architecture with TODOs

aman-goel:
This PR adds the first version of cleaned up explicit-state model checking framework for P, called PExplicit.
This version provides an initial architecture with several remaining TODOs to be completed in future PRs.
There are 2 parts:
- PExplicit IR: This is the new PCompiler backend that generates the IR dedicated for PExplicit.
- PExplicit RT: PExplicit runtime that will run the main explicit-state model checking engine.

Changes include:

[PExplicit IR]:
- Adds the first version of cleaned up basic IR code generator (triggered by hidden mode `--mode explicit`)

[PExplicit RT]:
- Adds PValues classes (like from PObserve) to encode P data types.
- Adds basic harnesses for PMachine, PMonitors, PEvent, etc.
- Adds basic architecture of Scheduler.
- Adds supporting utilities such as logging, exceptions, etc.


Src/PChecker/CheckerCore/CheckerMode.cs
```diff
@@ -16,5 +16,9 @@ public enum CheckerMode
     /// <summary>
     /// Mode for exhaustive explicit-state search with state-space coverage reporting
     /// </summary>
-    Coverage
+    Coverage,
```
ankushdesai, line 19: Can we drop coverage?

aman-goel: Let's drop this once we have a minimal running version. I wish to retain this for experimental comparison purposes. Note that this is hidden from P CLI.

ankushdesai:
Lets not merge these changes into master. Create a branch, dev/pexplicit_checker and merge the changes into that branch.

Src/PRuntimes/PExplicitRuntime/src/main/java/pexplicit/runtime/PExplicitGlobal.java
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * Status of the run
+     **/
+    @Getter
+    @Setter
+    private static String result = "error";
+    /**
+     * Result of the run
+     **/
+    private final Map<String, StateEvents> allStateEvents = new HashMap<>();    /* Mapping of each machine's state with its corresponding event handlers **/
```
ankushdesai, line 41: Why do we have this here?

ankushdesai (changes requested):

Src/PRuntimes/PExplicitRuntime/src/main/java/pexplicit/runtime/PExplicitGlobal.java
aman-goel: Good catch. This should be put at a different place, inside the State class object.
Will revise and confirm other fields in PExplicitGlobal class. Thanks!

ankushdesai (approved):

Src/PRuntimes/PExplicitRuntime/src/main/java/pexplicit/runtime/PExplicitGlobal.java
aman-goel: To be addressed in next PR.

PR #722, opened 2024-04-22
[PExplicit] Support liveness/deadlock checking, state-caching, stateful backtracking

aman-goel:
Updates include:

[Deadlock/liveness]
- Adds deadlock checking
- Adds basic liveness checking, based on whether a monitor is in a hot state at the end of a schedule

[State caching]
- Adds state caching to avoid revisiting same state more than once
- State caching mode can be: `none`, `fingerprint` (i.e., use hash codes), or `exact` (use exact values). Fingerprinting is the default.
- Adds hidden CLI option `--state-caching <mode>` (default is `fingerprint`)

[Stateful backtracking]
- Adds stateful backtracking to directly jump to the choice state instead of restarting from step zero at every iteration.
- Each unexplored choice stores the scheduler state at the start of the corresponding scheduler step.
- Stateful backtracking is completely optional and can be later tuned. If the unexplored choice selected in a new iteration has the stored state, stateful backtracking is done. If stored state is null, restarting from step zero is done.
- Adds hidden CLI option `--no-backtrack` to disable stateful backtracking (by default it is enabled).

[Replaying]
- Adds buggy trace replayer to confirm and replay the buggy trace when a bug is found.
- Adds basic `.schedule` writer when replaying (currently based on sender queue semantics)
- Adds TextWriter to write trace in `.txt` format when replaying (currently based on sender queue semantics)

[Misc]
- Updates logger to resemble like in C#
- Updates logging and stats to track and report distinct states
- Minor fixes, reformatting, and corrections


Src/PRuntimes/PExplicitRuntime/src/main/java/pexplicit/runtime/machine/MachineLocalState.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.io.Serializable;
+import java.util.List;
+
+/**
+ * Represents the local state of a machine
+ */
+public class MachineLocalState implements Serializable {
+    /**
+     * List of values of all local variables (including internal variables like currentState, FIFO queue, etc.)
```
ankushdesai, line 14: I am a bit worried if this is the best way to track local state.

Src/PRuntimes/PExplicitRuntime/src/main/java/pexplicit/values/ComputeHash.java
```diff
@@ -43,4 +45,18 @@ public static int getHashCode(PMachine machine, PValue<?>... values) {
         }
         return hashValue;
     }
+
+    /**
+     * Compute hash value for a PMachine local variables.
+     */
+    public static int getHashCode(SortedSet<PMachine> machines) {
```
ankushdesai, line 52: I am hoping this way of computing hash is prototype and we improve on this?

Src/PRuntimes/PExplicitRuntime/src/main/java/pexplicit/runtime/machine/PMachine.java
```diff
@@ -125,29 +120,122 @@ public boolean canRun() {
...
+        return result;
     }
 
     /**
-     * TODO
+     * Get values of local variables as an ordered list
+     *
+     * @return List of values
+     */
+    public List<Object> getLocalVarValues() {
```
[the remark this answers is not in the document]
aman-goel: This is the first version. As we develop this further, we can update it.

For state caching and stateful backtracking, we need a way to access all field values associated with the machine. These include all field values declared as the subclass that extends the `PMachine` abstract class (declared in corresponding PExplicit IR) as well as all fields inherited from the `PMachine`.

An alternative is to use [getDeclaredFields()](https://stackoverflow.com/questions/2172263/java-reflection-find-fields-of-a-subclass) Java function, but things become more complicated with monitors, making the inheritance to for e.g., `PMachine` -> `PMonitor` -> `Client`.

Any suggestions welcomed!

Src/PRuntimes/PExplicitRuntime/src/main/java/pexplicit/runtime/machine/MachineLocalState.java
aman-goel: Let's discuss this further offline and revamp this in the next PR.

Src/PRuntimes/PExplicitRuntime/src/main/java/pexplicit/values/ComputeHash.java
aman-goel: Yes, indeed. A much improved version of state caching to follow in the next PR.

ankushdesai (approved):

PR #723, opened 2024-04-24
PR cleanup

aoli-al:


Src/PChecker/CheckerCore/Coverage/CoverageInfo.cs
```diff
@@ -54,8 +54,6 @@ public CoverageInfo()
             Machines = new HashSet<string>();
             MachinesToStates = new Dictionary<string, HashSet<string>>();
             RegisteredEvents = new Dictionary<string, HashSet<string>>();
-            EventInfo = new EventCoverage();
-            CoverageGraph = new Graph();
```
ankushdesai, line 58: why this change?

Src/PChecker/CheckerCore/Coverage/ActorRuntimeLogEventCoverage.cs
ankushdesai, line 1: Can we remove this renaming?

ankushdesai (changes requested):

Src/PChecker/CheckerCore/Coverage/CoverageInfo.cs
aoli-al: This change was introduced because of merge. Removing this aligns with the upstream https://github.com/p-org/P/blob/master/Src/PChecker/CheckerCore/Coverage/CoverageInfo.cs#L56

Src/PChecker/CheckerCore/Coverage/ActorRuntimeLogEventCoverage.cs
aoli-al: Yes, this is only for `p-org:experimental/feedback-strategy`, if you compare this branch with master branch you won't see this change.

aoli-al:
I noticed that you have already merged part of this PR through #719 . Those commits are removed.

PR #725, opened 2024-05-02
Limit number of choices in choose expression to at most 10000

aman-goel:
Updates include:

[PCompiler]
- Throw a static error if N is integer literal greater than 10K in choose(N)

[C# and PSym Runtimes]
- Throw an error if N is integer variable greater than 10K in choose(N)
- Throw an error if N is seq/set/map with greater than 10K elements in choose(N)

[Docs]
- Update P GitHub manual for choose to reflect this change

[Tst]
- Adds new unit tests


ankushdesai (approved):
Thanks!

PR #745, opened 2024-06-14
[PEx] Revamps tracking unexplored choices, changes schedule choice

aman-goel:
Changes include:

[PEx Runtime]
- Adds `SearchUnit` class to track the list of unexplored choices at a given choice number
- Changes `SearchTask` class to now track schedule and `SearchUnit`s separately
- Track schedule to be just a list of current choices
- Eliminates unnecessary copy-pasting of unexplored choices from `SearchTask` to `Schedule`.- Changes schedule choice to be a `PMachineID` instead of a `PMachine`


Src/PRuntimes/PExplicitRuntime/src/main/java/pexplicit/runtime/PExplicitGlobal.java
```diff
@@ -60,19 +61,18 @@ public class PExplicitGlobal {
...
-     * @param idx  Machine index
+     * @param pid Machine pid
      * @return Machine
      */
-    public static PMachine getGlobalMachine(Class<? extends PMachine> type, int idx) {
-        List<PMachine> machinesOfType = machineListByType.get(type);
+    public static PMachine getGlobalMachine(PMachineId pid) {
+        List<PMachine> machinesOfType = machineListByType.get(pid.getType());
         if (machinesOfType == null) {
             return null;
```
ankushdesai, line 70: When would this be null?

Src/PRuntimes/PExplicitRuntime/src/main/java/pexplicit/runtime/PExplicitGlobal.java
```diff
@@ -60,19 +61,18 @@ public class PExplicitGlobal {
...
         }
-        if (idx >= machinesOfType.size()) {
+        if (pid.getTypeId() >= machinesOfType.size()) {
             return null;
```
ankushdesai, line 73: When would this be the case? Should we instead throw an exception?

Src/PRuntimes/PExplicitRuntime/src/main/java/pexplicit/runtime/PExplicitGlobal.java
aman-goel: When we are allocating the machine the very first time [here](https://github.com/p-org/P/blob/dev/pexplicit_checker/Src/PRuntimes/PExplicitRuntime/src/main/java/pexplicit/runtime/scheduler/Scheduler.java#L276)

Src/PRuntimes/PExplicitRuntime/src/main/java/pexplicit/runtime/PExplicitGlobal.java
aman-goel: When we are allocating a machine instance the very first time ([here](https://github.com/p-org/P/blob/dev/pexplicit_checker/Src/PRuntimes/PExplicitRuntime/src/main/java/pexplicit/runtime/scheduler/Scheduler.java#L276)), this function can return null. In that case, we create the machine instance and add it to global context ([here](https://github.com/p-org/P/blob/dev/pexplicit_checker/Src/PRuntimes/PExplicitRuntime/src/main/java/pexplicit/runtime/scheduler/Scheduler.java#L277-L280)).

PR #748, opened 2024-07-09
Dev/ashish

xashisk:
Extending PeX for multiple threads within a single core implementing dynamic partitioning of the execution tree. Subtasks completed include:
        1. Pulled out first createTask method into RuntimeExecutor in process() before submitting the jobs to the executor object, but after making the threadcopies. 
        2. Made createTask also static in addition to createFirstTask. Also changed all function calls to this function to call it in a static way.
        3. In ExplicitSearchScheduler made a runParallel() method. In the Scheduler class added prototype for abstract runParallel (but let run() prototype still be there.). Finally in TimedCall Object, call the runParallel() from scheduler object.
        4. SearchStrategy.java - allTasks, pendingTasks, finishedTasks - made these thread safe by using the Collections.synchonizedSet (for HashSet) and Collections.synchonizedList (for ArrayList).
        5. machineListByType in PExplicitGlobal is made per thread - made a hashmap from local threadId (a.k.a. thread index) to machineListByType. For the original lombak Getter for machineListByType, implemented a public getMachineListByType function in PExplicitGlobal. Within the getGlobalMachine and addGlobalMachine, made a local copy of machineListByType by calling getMachineListByType.  
        6. stateCache in ExplicitSearchScheduler.java - made it a ConcurrentHashMap object. 
        7. Made a threadSafe global variable to count number of currently Blocked Threads. Within setNextTask() in SearchStrategy.java implemented a feature to block a thread if pendingTasks is Empty, and keep blocking till pendingTasks is Empty; unless all threads are blocking here in which case, all threads just return. 
        8. In PExplicitLogger, replaced all calls to log.Info by the logInfo function (except the call to log.Info inside logInfo itself). 
        9. ElementList and elementSet from SearchStrategyRandom and SearchStrategyAstar - made it static and synchronized versions.
        10. In getMachineListByType() in PExplicitGlobal, if the key is not there, initialize that value to an empty Hashmap and then return this newly initialize HashMap.
        11. Added a numThreads option to PExplicitConfig, with default value set to 1. Set the maxThreads variable in PExplicitGlobal from this numThreads variable. Also added a CLI option in PExplicitOptions (at line 151) with a single variable of Integer type. Named the command line argument ‘nproc’. Also added a case ‘nproc’ in the switch statement in ParseCommandlineArgs, which initializes the numThreads in PExplicitConfig.
        12. In PExplicitGlobal, made machineSet variable per thread (changed it’s name) by using a ConcurrentHashMap; and added a getMachineSet function interface for it.
        13.  Changed all log.info libary calls in PExplicitLogger to logInfo() procedure call.
        14. Made separate file for thread based logging called PExplicitThreadLogger. Within this file changed from console based logging to file based logging.
        15. All log calls inside runParallel are made to PExplicitThreadLogger, and log calls outside runParallel are made to PExplicitLogger. Added a call to PExplicitThreadLogger.Initialize() in order to create the file.
        16. Ran and debugged on sample projects, finished milestone A.1


aman-goel:
@xashisk Can you resolve the merge conflicts? Thanks!

PR #749, opened 2024-07-12
[PEx] Adds hybrid stateful backtracking, improves storing search tasks

aman-goel:
Updates include:

[PEx RT]
- Serialize/deserialize memory-heavy fields of `SearchTask` as needed since taking too much space for long runs
- Added hydrid stateful backtracking that performs stateful backtracking for search within a search task only
- Adds CLI option `--stateful-backtrack <none|intra-task|all>` (default is `intra-task`). Deprecates option `--no-backtrack`
- Minor refactoring


ankushdesai (approved):
Looks good!

PR #754, opened 2024-07-25
Fixing "key not found" error when using set as key

ChristineZh0u:
Fixing `key not found` error thrown when user does this:
```
var foomap: map[set[int], int];
foomap[default(set[int])] = 0;
print format("{0}", foomap[default(set[int])]);
```


ChristineZh0u:
The failed test case "tcMapSet" is replicating an issue that is fixed by this commit. The Tutorials CI is meant for latest released version. Since this change fixes a language-level buggy behavior, the Tutorials is expected to fail.

ankushdesai (approved):

PR #757, opened 2024-08-14
Merging Actor.cs with StateMachine.cs

ChristineZh0u:


Src/PChecker/CheckerCore/Actors/ActorFactory.cs
```diff
@@ -15,17 +15,17 @@ internal static class ActorFactory
         /// <summary>
         /// Cache storing actors constructors.
```
ankushdesai, line 16: Replace actors in the comments as well with state machine

ankushdesai (changes requested):
Can you please make sure that the comments are also updated to remove references to actor?

ankushdesai (approved):

PR #760, opened 2024-08-16
[Cleanup] Moving ActorRuntime into ControlledRuntime

ChristineZh0u:
- Renamed ActorRuntime log file names to PChecker logs
- Moved ActorRuntime into ControlledRuntime


ankushdesai (approved):
Thanks!

PR #766, opened 2024-08-28
[Cleanup] Merging PMonitor with Monitor

ChristineZh0u:
Test:
- Made sure all unit tests passed
- Checked logs are generating correctly


Src/PChecker/CheckerCore/Specifications/Monitor.cs
```diff
@@ -246,6 +297,16 @@ protected void RaiseGotoStateEvent(Type state)
             CheckDanglingTransition();
             PendingTransition = new Transition(Transition.Type.Goto, state, default);
         }
+        
+        public void TryAssert(bool predicate)
```
ankushdesai, line 301: Do we need this function? Instead, the generated code should directly call `Assert` function that this function is calling.

Src/PChecker/CheckerCore/Specifications/Monitor.cs
```diff
@@ -246,6 +297,16 @@ protected void RaiseGotoStateEvent(Type state)
...
+        {
+            Assert(predicate);
+        }
+
+        public void TryAssert(bool predicate, string s, params object[] args)
```
ankushdesai, line 306: Same as above.

Src/PCompiler/CompilerCore/Backend/CSharp/CSharpCodeGenerator.cs
```diff
@@ -263,7 +263,7 @@ private void WriteMonitor(CompilationContext context, StringWriter output, Machi
             WriteNameSpacePrologue(context, output);
 
             var declName = context.Names.GetNameForDecl(machine);
-            context.WriteLine(output, $"internal partial class {declName} : PMonitor");
+            context.WriteLine(output, $"internal partial class {declName} : PChecker.Specifications.Monitors.Monitor");
```
ankushdesai, line 266: Instead of this; add `using PChecker.Specifications.Monitors;` and then just have the `Monitor`

Src/PCompiler/CompilerCore/Backend/CSharp/CSharpCodeGenerator.cs
```diff
@@ -737,7 +737,7 @@ private void WriteFunction(CompilationContext context, StringWriter output, Func
                 // for monitor
                 if (!(function.CanCreate == true || function.CanSend == true || function.IsNondeterministic == true || function.CanReceive == true))
                 {
-                    var functionParameters_monitor = functionParameters + string.Concat(seperator, "PMonitor currentMachine");
+                    var functionParameters_monitor = functionParameters + string.Concat(seperator, "PChecker.Specifications.Monitors.Monitor currentMachine");
```
ankushdesai, line 740: Same as above;

Src/PChecker/CheckerCore/Specifications/Monitor.cs
```diff
@@ -177,6 +187,17 @@ internal void Initialize(ControlledRuntime runtime)
...
             Runtime = runtime;
         }
+        
+        /// <summary>
+        /// Checks if the event is null, and calls <see cref="Monitor.RaiseEvent"/>.
+        /// </summary>
+        public void TryRaiseEvent(Event ev, object payload = null)
+        {
+            Assert(!(ev is DefaultEvent), "Monitor cannot raise a null event");
+            var oneArgConstructor = ev.GetType().GetConstructors().First(x => x.GetParameters().Length > 0);
```
ankushdesai, line 197: Do we need this part? Can we not set the `payload` field in the `ev` in the generated code and then just call `RaiseEvent` directly.

Src/PChecker/CheckerCore/Specifications/Monitor.cs
```diff
@@ -196,6 +217,36 @@ protected void RaiseEvent(Event e)
...
+            Logger.Write($"{message}");
+        }
+        
+        /// <summary>
+        /// Stores payload temperarily, and calls <see cref="Monitor.RaiseGotoStateEvent"/>.
+        /// </summary>
+        public void TryGotoState<T>(object payload = null) where T : State
+        {
+            gotoPayload = payload;
+            RaiseGotoStateEvent<T>();
```
ankushdesai, line 247: Change the `RaiseGotoStateEvent` to take a parameter `payload` and then in that function set the gotoPayload field. Then eliminate the TryGotoState function.

ankushdesai (changes requested):

Src/PChecker/CheckerCore/Specifications/Monitor.cs
ChristineZh0u: Calling assert function directly results in this error: `'Assert(bool)' is inaccessible due to its protection level`, when I try setting Assert function to public, it results in: `'StateMachine.Assert(bool)' is inaccessible due to its protection level`. Is it ok if all assert functions are set to public?

ankushdesai: Yes.

PR #770, opened 2024-09-05
[Cleanup] Merging PMachine with StateMachine

ChristineZh0u:
- Removed InitializeParametersFunction(), class InitializeParameters, and class InitializeParametersEvent
- Removed most Try..() functions
- Made sure all unit test passes and generated logs are correct


Src/PChecker/CheckerCore/SystematicTesting/ControlledRuntime.cs
```diff
@@ -457,7 +460,14 @@ private StateMachine CreateStateMachine(StateMachineId id, Type type, string nam
 
             if (id is null)
             {
-                id = new StateMachineId(type, name, this);
+                if (name != null)
+                {
+                    id = new StateMachineId(type, name.Substring(2), this);
```
ankushdesai, line 465: What is this?

ankushdesai (changes requested):

Src/PChecker/CheckerCore/SystematicTesting/ControlledRuntime.cs
ChristineZh0u: This is because the new state machine's `InterfaceName` couldn't be passed in through `InitializeParametersEvent` anymore. I decided to pass it in is through the `name` parameters in CreateStateMachine() since it's not used for anything else. This function is originally getting called like this:
`var machineId = CreateStateMachine(createMachine, createdInterface.Substring(2), new InitializeParametersEvent(new InitializeParameters(createdInterface, payload)));` 
Now it gets called like this:
`var machineId = CreateStateMachine(createMachine, createdInterface, GetConstructorEvent(payload));`

ankushdesai (approved):

PR #789, opened 2024-10-08
Removing compiler changes

ChristineZh0u:
Creating separate PRs for each cleanup task. This PR removes the changes in PCompiler


aoli-al:
Can you double-check the changes we made in the CompilerCore? https://github.com/p-org/P/compare/master...experimental/feedback

You reverted more changes than needed.

ChristineZh0u:
> Can you double-check the changes we made in the CompilerCore? [master...experimental/feedback](https://github.com/p-org/P/compare/master...experimental/feedback)
> 
> You reverted more changes than needed.

The reason `PEvent.cs`, `Event.cs`, and `PMachine.cs` are changed is because there's a variable in `CSharpCodeGenerator.cs` called `_sendEventIndex` which is uses as a parameter called `int Loc` when initializing an Event. Is that variable needed? Or did I removed too much from somewhere else?

aoli-al:
For example, I don't think I have changed `.github/workflows/pcover.yml`, which is included in your change list. I'm working on a rebuttal these days and will take a closer look this Friday. 

A good way to verify is to check if you have changed more files listed here https://github.com/p-org/P/compare/master...experimental/feedback

ChristineZh0u:
> For example, I don't think I have changed `.github/workflows/pcover.yml`, which is included in your change list. I'm working on a rebuttal these days and will take a closer look this Friday.
> 
> A good way to verify is to check if you have changed more files listed here [master...experimental/feedback](https://github.com/p-org/P/compare/master...experimental/feedback)

Ohh i see, all the `.github` folder changes are to eliminate the github warnings. I'll revert them back. And yes, I've made sure that all changes are in the https://github.com/p-org/P/compare/master...experimental/feedback. Good luck with rebuttal!

ankushdesai (approved):
Nice!

PR #799, opened 2024-10-18
Added VectorTime and BehavioralObserver class for feedback strategy

ChristineZh0u:
Currently the BehavioralObserver calculates the uniqueness score for each schedule's timeline and prints the score out in the console. Next step is to integrate/replace the TimelineObserver with BehavioralObserver. 

The uniqueness score is calculated through MinHash. The function randomly chooses 100 elements (each element is a tuple of (vector time, event, SEND/DEQUEUE enum)) from the current timeline and compare them with all past timelines. The final score is the average number of miss-matches between the current timeline and the past timelines. If there is a duplicate timeline int the past, the uniqueness score will be 0.


ankushdesai (approved):
This is fantastic work!! I am super impressed with this PR!

PR #805, opened 2024-12-03
Start to move PVerifier so that its ready to be merged into P3.0

ankushdesai:


Src/PCompiler/CompilerCore/Parser/PLexer.g4
```diff
@@ -109,6 +129,7 @@ NONDET     : '$'  ;
 LNOT   : '!' ;
 LAND   : '&&' ;
 LOR    : '||' ;
+LTHEN  : '==>';
```
ankushdesai, line 132: Do you think we can add `<==>` for `IFF` or add `IFF` as a keyword.

Src/PCompiler/CompilerCore/TypeChecker/AST/Declarations/Function.cs
```diff
@@ -70,7 +75,15 @@ public void AddLocalVariable(Variable local)
         {
             localVariables.Add(local);
         }
-
+        public void AddRequire(IPExpr expr)
```
ankushdesai, line 78: I remember there were no issues with pure functions where a pure function must not refer to a machine field. Can we add a static check for this?

Src/PCompiler/CompilerCore/TypeChecker/AST/Expressions/TestExpr.cs
```diff
@@ -0,0 +1,25 @@
+using System.Diagnostics;
+using Antlr4.Runtime;
+using Plang.Compiler.TypeChecker.AST.Declarations;
+using Plang.Compiler.TypeChecker.Types;
+
+namespace Plang.Compiler.TypeChecker.AST.Expressions
```
ankushdesai, line 6: Can you remind me what is this for?

ankushdesai: Why has this changed?

ankushdesai: Thanks! Can we rename this to `TestADTExpr` to make this clearer? Thanks!

PR #809, opened 2024-12-15
Java code generation fixes

ChristineZh0u:
- Reverted PEvent Java code generation bug introduced in P Checker clean up
- Added the JEvent case to the existing handling of immutable types (like JString and JEnum)
- Changed the default value for JEvent to be null


lewisbru:
This looks correct (per my local changes)

ankushdesai (approved):

PR #815, opened 2025-01-23
Added test for generated PChecker logs

ChristineZh0u:
Added test for PChecker log generation against p model: Feature1SMLevelDecls/DynamicError/bug2/bug2.p. 
Also removed "timeline.txt" generation since nothing gets written into this file currently.

### DIFF
--- a/Src/PChecker/CheckerCore/SystematicTesting/TestingEngine.cs
+++ b/Src/PChecker/CheckerCore/SystematicTesting/TestingEngine.cs
@@ -165,9 +165,7 @@ namespace PChecker.SystematicTesting
         /// A guard for printing info.
         /// </summary>
         private int PrintGuard;
-
-        private StreamWriter TimelineFileStream;
-
+        
 
         /// <summary>
         /// Creates a new systematic testing engine.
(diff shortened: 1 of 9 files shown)


Src/PChecker/CheckerCore/SystematicTesting/TestingEngine.cs
```diff
@@ -272,7 +270,6 @@ private TestingEngine(CheckerConfiguration checkerConfiguration, TestMethodInfo
 
             CancellationTokenSource = new CancellationTokenSource();
             PrintGuard = 1;
-            TimelineFileStream = new StreamWriter(checkerConfiguration.OutputDirectory + "timeline.txt");
```
lewisbru, line 275: Is this being removed because it was causing the new test to fail (because of the unexpected file)?  Or is it just cleanup?

ChristineZh0u: Both. First it was causing the new test to fail. Then during debugging, I found that nothing is writing into the stream.

ankushdesai (approved):

PR #831, opened 2025-03-11
Added better error messages for assertion and ran auto code cleanup.

ankushdesai:


Src/PChecker/CheckerCore/ExhaustiveEngine.cs
```diff
@@ -148,19 +148,19 @@ private async Task CreateTaskFromProcess(Process proc)
                 switch (proc.ExitCode)
                 {
                     case 0:
-                        Logger.WriteLine($"... Checker run finished.");
+                        _logger.WriteLine("... Checker run finished.");
```
ChristineZh0u, line 151: In future code, should we format all private fields to be: `_fieldName`?

Src/PChecker/CheckerCore/Runtime/StateMachines/StateMachine.cs
```diff
@@ -534,12 +533,12 @@ protected StateMachineId CreateStateMachine(Type type, string name, Event initia
         /// <param name="opGroupId">Optional id that can be used to identify this operation.</param>
         public void SendEvent(PMachineValue target, Event ev)
         {
-            Assert(ev != null, "Machine cannot send a null event");
-            Assert(target != null, "Machine in send cannot be null");
+            Assert(ev != null, $"Machine cannot send a null event. Machine {Id} trying to send null event in state {CurrentStateName}.");
```
ChristineZh0u, line 536: Don't think this assertion will ever be false. When I try to send a null event it gets caught by the compiler.

ChristineZh0u (approved):

PR #833, opened 2025-03-13
Param

zhezhouzz:
# Parametric Testing

This document introduce the parametric tesing feature.

## Syntax

We introduce the following new syntax to P language.

```
topDecl:                # Top-level P Program Declarations
...
| ParamDecl             # The parameters declaration
| ParamTestDecl         # The parametric test case declaration 
;

ParamDecl:
| param iden: type      # Support bool, int, float types.
;

PrimitiveSet :          # A nonempty set of primitive expressions (can be added as set literal into primitiveExpr)
| primitiveExpr (, primitiveExpr)*
; 

ParamAssignment:        # A nonempty set of assignments
| iden in [ PrimitiveSet ] (, iden in [ PrimitiveSet ])*
; 

ParamTestDecl:
| paramtest (ParamAssignment) assume (expr) test iden [main=iden] : modExpr ;
;
```

The parameters are global variables that can be referred to by all P machines or functions. During the compilation of each parametric test case, the P compiler will instantiate these parameters according to the provided assignments (e.g., a set of primitive values) of these parameters (i.e., `paramtest`). One single parameterized test will be compiled into multiple concrete test cases, which include all possible combinations of parameter assignments that are allowed by the assumption expression.

## Examples and Semantics

A parametric test case `paramTestCase` is defined as follows, where machines `M1` and `M2` do not refer to the global parameter `p4`. Then the test case `paramTestCase` just needs to provide the assignment of `p1`, `p2`, and `p3`.

```
param p1: int;
param p2: int;
param p3 : bool;

paramtest (p1 in [1,5], p2 in [3, 2], p3 = [true]) assume (p1 > p2) paramTestCase [main=Test]:
  assert pspec in (union M1 M2);
```

Without the constraints from the assumption, there are `4` possible combinations. However, the assumption `p1 > p2` rejects the first two, so the compiler will only instantiate two concrete test cases:

```
p1 := 1, p2 = 3, p3 = true   # inconsistent with assumption
p1 := 1, p2 = 2, p3 = true   # inconsistent with assumption
p1 := 5, p2 = 3, p3 = true
p1 := 5, p2 = 2, p3 = true
```

The instantiated test cases are renamed according to the assignment combination. For example, the compiled test cases in this example are:

```
paramTestCase_p1_5_p2_3_p3_true
paramTestCase_p1_5_p2_2_p3_true
```

The compiled test cases are semantically equal to replacing all referred parameters with the assigned primitive values. In this example, it is equal to the following two normal test cases (`[ x -> v]` indidates substitue vairable `x` with value `v`):

```
test paramTestCase_p1_5_p2_3_p3_true [main=Test]:
  assert pspec[p1 -> 5, p2 -> 3, p3 -> true] in 
  (union M1 M2)[p1 -> 5, p2 -> 3, p3 -> true];

test paramTestCase_p1_5_p2_2_p3_true [main=Test]:
  assert pspec[p1 -> 5, p2 -> 2, p3 -> true] in 
  (union M1 M2)[p1 -> 5, p2 -> 2, p3 -> true];
```

### Assumption

The assumption will be checked during compile time by a simple interpreter of P expressions, where we support basic arithmetic operators over `int` and `bool`. 

### Backend

There will be a new static class `GlobalConfig` under the `PImplementation` namespace, which contains the declaration of all parameters. These parameters are initialized following P's initialization manner. For example, assuming the P code contains `4` parameters `g1`, `g2`, `nClients`, and `b1`, the compiled C# will include the following class:

```C#
namespace PImplementation
{
    public static class GlobalConfig
    {
          public static PInt g2 = ((PInt)0);
          public static PInt g1 = ((PInt)0);
          public static PInt nClients = ((PInt)0);
          public static PBool b1 = ((PBool)false);
    }
}
```

The compiled P machine will refer to these parameters as if referring to other variables. Each compiled test case will initialize these parameters according to their assignments via a function `InitializeGlobalConstantVariables`. For example:

```C#
namespace PImplementation
{
    public class aaaa1___nClients_2__g1_1__g2_4 {
        public static void InitializeGlobalConstantVariables() {
              (GlobalConfig.nClients) = ((PInt)(2));
              (GlobalConfig.g1) = ((PInt)(1));
              (GlobalConfig.g2) = ((PInt)(4));
        }
...
        [PChecker.SystematicTesting.Test]
        public static void Execute(ControlledRuntime runtime) {
            InitializeGlobalConstantVariables();
...
```

In this example, the test case `aaaa1` has the assignment `[nClients -> 2, g1 -> 1, g2 -> 4]`, and it will initialize these parameters before each execution. Variables without assignments (e.g., `b1`) will retain their initial values.

Here's the grammar-checked version:

## Changes (Compare with P 3.0)

[The files changed between this branch and major/P3.0](https://github.com/p-org/P/compare/major/P3.0...zhezhouzz:P:param)

+ Backends (`Src/PCompiler/CompilerCore/Backend/`), mainly change the file `PCheckerCodeGenerator.cs`:
  + Add a wrapper function `GetGlobalParamAndLocalVariableName` to refer to both parameters and local variables.
  + Add a function `DeclareGlobalParams` that outputs a C# function that initializes parameters.
  + Modify the function `WriteSafetyTestDecl` to handle both normal tests and parametric tests. There is a loop to calculate all possible assignment `index` combinations for parametric tests. There is a simple interpreter `Eval` to check the assumption.
  + Modify the function `WriteTestFunction` to take an additional parameter which decides if it is a parametric test. If it is, call the parameter initialization function.
+ Error Handling (`Src/PCompiler/CompilerCore/DefaultTranslationErrorHandler.cs` and `Src/PCompiler/CompilerCore/ITranslationErrorHandler.cs`):
  + Add three new type errors: `RedeclareGlobalParam`, `UndeclaredGlobalParam`, and `ModifyGlobalParam`.
+ Syntax (`Src/PCompiler/CompilerCore/Parser/PLexer.g4` and `Src/PCompiler/CompilerCore/Parser/PParser.g4`), following the syntax section above.
+ AST definition (`Src/PCompiler/CompilerCore/TypeChecker/AST`):
  + File `SafetyTest.cs` adds three types of tests: normal, parametric without assumption, parametric with assumption.
  + Files `Expressions/BoolLiteralExpr.cs` and `Expressions/IntLiteralExpr.cs` allow construction of these objects from bool or int values (for assumption interpreter).
  + File `Expressions/SeqLiteralExpr.cs` implements sequence literals.
+ AST Visitor (`Src/PCompiler/CompilerCore/TypeChecker/`) builds corresponding visitors for new syntax.
+ Type check (`Src/PCompiler/CompilerCore/TypeChecker/Scope.cs`).
+ Some test cases (`Tst/TestParamTester`).


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "Ext/libhandler"]
```
ankushdesai, line 1: why are we adding these back? Can we remove all these unrelated changes?

Src/PCompiler/CompilerCore/TypeChecker/AST/Declarations/SafetyTest.cs
```diff
@@ -1,6 +1,12 @@
 ﻿using Antlr4.Runtime;
 using System.Collections.Generic;
 
+public enum TestKind {
+    NormalTest,
+    ParametricTest,
+    AssumeParametricTest,
```
ankushdesai, line 7: Can we merge the Parametric and AssumeParametric Tests?

Src/PCompiler/CompilerCore/TypeChecker/ModuleExprVisitor.cs
```diff
@@ -13,12 +13,38 @@ namespace Plang.Compiler.TypeChecker
 {
     internal static class ModuleSystemDeclarations
     {
+
+        public static IDictionary<string, List<IPExpr>> make_assignments(ITranslationErrorHandler handler, Scope globalScope,
```
ankushdesai, line 17: Can we give a better name, make assignments to what?

Src/PCompiler/CompilerCore/TypeChecker/ModuleExprVisitor.cs
```diff
@@ -30,34 +56,37 @@ public static void PopulateAllModuleExprs(
...
+                        
+                    }
+                    case TestKind.ParametricTest:
+                    {
+                        var context = (PParser.ParametricSafetyTestDeclContext)test.SourceLocation;
+                        test.ModExpr = modExprVisitor.Visit(context.modExpr());
+                        var expr = (NamedTupleExpr)paramExprVisitor.Visit(context.globalParam);
+                        var dic = make_assignments (handler, globalScope, context.globalParam, expr);
+                        test.ParamExpr = dic;
+                        // the default assumption is true
```
ankushdesai, line 85: I have a strong feeling we can merge the AssumeParam and NormalParam together; making NormalParam a special case of AssumeParam where the assumes are all true?

ankushdesai (changes requested):

Src/PCompiler/CompilerCore/Parser/PParser.g4
```diff
@@ -255,6 +255,7 @@ param : LPAREN paramBody RPAREN;
 
 testDecl : TEST testName=iden (LBRACK MAIN ASSIGN mainMachine=iden RBRACK) COLON modExpr SEMI                  # SafetyTestDecl
          | PARAMTEST globalParam=param testName=iden (LBRACK MAIN ASSIGN mainMachine=iden RBRACK) COLON modExpr SEMI # ParametricSafetyTestDecl
+         | PARAMTEST globalParam=param ASSUME assumeExpr=expr testName=iden (LBRACK MAIN ASSIGN mainMachine=iden RBRACK) COLON modExpr SEMI # ParametricAssumeSafetyTestDecl
```
ankushdesai, line 258: Do we need to add another rule or can we also merge the two param test kinds making the assume optional?

ankushdesai (changes requested):
left one comment

ankushdesai (approved):

ankushdesai (approved):

PR #837, opened 2025-03-27
T-wise combinatorial test

zhezhouzz:
## T-wise Combinatorial Test

As discussed, add `t-wise` test. By default, it is `n`-wise where `n` is the number of parameters that are initialized in a safety test declaration, which means test all possible combinations. The `t` is supposed between `2` to `n`.

## Syntax

```
safetyTest:
 ... (pairwise | intLiteral wise)? testName=iden ...
```

User can put t-wise option before the name of test, `pairwise` is equal to `(2 wise)`.

## Change Log

+ new syntax (`Src/PCompiler/CompilerCore/Parser/PLexer.g4`, `Src/PCompiler/CompilerCore/Parser/PParser.g4`, `Src/PCompiler/CompilerCore/TypeChecker/AST/Declarations/SafetyTest.cs`).
+ new error (`Src/PCompiler/CompilerCore/DefaultTranslationErrorHandler.cs`, `Src/PCompiler/CompilerCore/ITranslationErrorHandler.cs`) when input `t` less than 2 or greater than the number of parameters.
+  Conbinatorial test (`Src/PCompiler/CompilerCore/TypeChecker/ParamAssignment.cs`).
+ examples (`Tst/TestParamTester/1_ClientServer/PTst/Testscript.p`)


Src/PCompiler/CompilerCore/DefaultTranslationErrorHandler.cs
```diff
@@ -71,6 +72,12 @@ public Exception ModifyGlobalParam(ParserRuleContext location, IPDecl existing)
                 $"try to modify a global param '{existing.Name}' at {locationResolver.GetLocation(existing.SourceLocation)}");
         }
 
+        public Exception InvalidTwise(ParserRuleContext location, IPDecl testDecl, int twiseNum, int paramNum)
+        {
+            return IssueError(location,
+                $"invalid twise: wise-number({twiseNum}) should great equal than 2 and less equal than number of parameters ({paramNum}) at {locationResolver.GetLocation(testDecl.SourceLocation)}");
```
lewisbru, line 78: In general, in t-wise combinatorial tests, t can be specified as [1,n] not [2, n].

Src/PCompiler/CompilerCore/TypeChecker/ModuleExprVisitor.cs
```diff
@@ -66,6 +66,26 @@ public static void PopulateAllModuleExprs(
...
+                }
+                else
+                {
+                    if (context.twise().PAIRWISE() != null)
+                    {
+                        test.Twise = 2;
+                    } else if (context.twise().WISE() != null)
+                    {
+                        var t = int.Parse(context.twise().IntLiteral().GetText());
+                        if (!(2 <= t && t <= test.ParamExprMap.Count))
```
lewisbru, line 82: Apart from whether t should be allowed to be 1, align this check with the error message (or vice-versa).

Src/PCompiler/CompilerCore/TypeChecker/ParamAssignment.cs
```diff
@@ -56,15 +161,35 @@ public static void IterateAssignments(SafetyTest safety, List<Variable> globalPa
...
+                continue;
+            }
+            assignmentIndices.Add(indexDic);
+            // Console.WriteLine($"Sat Assumption: indexArr: {string.Join(',', indexArr)}");
         } while (Next(indexArr, safety.ParamExprMap));
+        
+        foreach (var i in GreedyCoverageExplore(universe, assignmentIndices, safety.Twise))
+        {
+            // Console.WriteLine($"Choose {safety.Twise}-wise: indexArr: {string.Join(',', assignmentIndices[i])}");
+            f(IndexDic2Dic(globalParams, safety.ParamExprMap, assignmentIndices[i]));
```
lewisbru, line 146: I guess you had been using "f" as the name of the parameter for doing the action, but can you use a more descriptive name?

Src/PCompiler/CompilerCore/TypeChecker/ParamAssignment.cs
```diff
@@ -1,13 +1,118 @@
...
+        return assignmentCoverageMap;
+    }
+
+    private static List<int> GreedyCoverageExplore(Dictionary<string, int> universe, List<Dictionary<string, int>> assignmentList, int twise)
+    {
+        var assignments = Enumerable.Range(0, assignmentList.Count).ToList();
+        if (twise == universe.Count) return assignments;
+        var vectorMap = MakeVectorMap(universe, twise);
+        var assignmentMap = EnumerableToIndexDict(assignmentList, x => x);
+        // Console.WriteLine($"twise({twise})");
```
lewisbru: Do we want to have the various debugging trace code commented out?  If we think it will be useful later on, I think a better approach is to move it under a flag so that it can be used more easily and doesn't diverge from the code; otherwise, delete it.

Src/PCompiler/CompilerCore/TypeChecker/ParamAssignment.cs
```diff
@@ -1,13 +1,118 @@
...
+using System.Threading;
 using Plang.Compiler.TypeChecker.AST;
 using Plang.Compiler.TypeChecker.AST.Declarations;
 
 namespace Plang.Compiler.TypeChecker;
 
 public abstract class ParamAssignment
 {
+    
+    public static IEnumerable<IEnumerable<T>> DifferentCombinations<T>(IEnumerable<T> elements, int k)
```
lewisbru: I know there are a few end-to-end script tests, but given the complexity of this logic I suggest having some unit tests which actually verify the results of this code.

Src/PCompiler/CompilerCore/TypeChecker/ParamAssignment.cs
```diff
@@ -1,13 +1,118 @@
...
+        return k == 0 ? [Array.Empty<T>()]
+            :
+            enumerable.SelectMany((e, i) =>
+                DifferentCombinations(enumerable.Skip(i + 1), k - 1).Select(c => (new[] {e}).Concat(c)));
+    }
+    
+    private static Dictionary<int, T> EnumerableToIndexDict<T> (IEnumerable<T> l, Func<T, T> f)
+    {
+        return Enumerable.Range(0, l.Count()).Zip(l, (a, b) => new { a, b }).ToDictionary(x => x.a, x => f(x.b));
+    }
```
lewisbru, line 23: Nit:  extra line after the close brace.

Src/PCompiler/CompilerCore/DefaultTranslationErrorHandler.cs
zhezhouzz: We can always provide more options to the users, I will additional allow 1-wise test.

Src/PCompiler/CompilerCore/TypeChecker/ParamAssignment.cs
zhezhouzz: I have deleted these debugging code in the following commit.

PR #839, opened 2025-04-05
Implement SURW algorithm

aoli-al:
Implements the SURW algorithm with the following modifications:
- Treats each machine as a memory location
- Considers all events sent to a machine as interesting events
- Randomly samples 10% of machines as interesting machines

Testing notes:
- Translated paper examples into P models. The higher frequency of 0 values occurs because P reschedules on machine creation, increasing the probability of reads occurring before writes.
- Tests are disabled in CI due to limitation with loading multiple assemblies of the same name (conflict with `PCheckerLogGeneratorTests`). They can be run locally.

log_distribution


Shabba906 (approved):

Shabba906 (commented):
checkercore

PR #852, opened 2025-04-28
[FIX] Add sanity check to default

Ang9876:
1. Add sanity checks (received is a subset of sent) as preconditions to each procedure.
2. Extend `default` with sanity checks.
3. Fix the missing remaining invariants (`LEMMA`) when `prove default using LEMMA`
4. Update received buffer after inlined procedure.

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/PVerifier/Uclid5CodeGenerator.cs
+++ b/Src/PCompiler/CompilerCore/Backend/PVerifier/Uclid5CodeGenerator.cs
@@ -38,6 +38,7 @@ public class PVerifierCodeGenerator : ICodeGenerator
     private Dictionary<Invariant, HashSet<Invariant>> _invariantDependencies;
     private HashSet<Invariant> _provenInvariants;
     private Scope _globalScope;
+    private Invariant _defaultInv;
 
     public bool HasCompilationStage => true;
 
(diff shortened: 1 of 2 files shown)


Src/PCompiler/CompilerCore/Backend/PVerifier/Uclid5CodeGenerator.cs
```diff
@@ -424,29 +438,26 @@ private List<CompiledFile> CompileToFile(string name, ProofCommand cmd)
...
-                        }
+                        _src = GenerateCompiledFile(cmd, name, m, s, null);
+                        files.Add(_src);
+                        GenerateMain(m, s, null, cmd, cmd.Name == "default");
+                    }
 
-                        foreach (var e in events.Where(e => !e.IsNullEvent && !e.IsHaltEvent && s.HasHandler(e)))
+                    foreach (var e in events.Where(e => !e.IsNullEvent && !e.IsHaltEvent && s.HasHandler(e)))
+                    {
+                        if (_ctx.Job.CheckOnly == null || _ctx.Job.CheckOnly.Contains(e.Name))
```
AD1024, line 456: What is the semantic of `CheckOnly` here? Seems we are checking both machine/state and event names. This might not work if I want to check all event handlers for a single machine (suppose I pass only the name of the machine to `CheckOnly`).

AD1024: @Ang9876

Ang9876: I did not change this part of code but only remove `else`. My understanding is that the part corresponding to `machine` generates a file for `entry`, and that corresponding to `events` generates files for event handlers. So I guess there is no way to specify a single machine and check all event handlers (without checking event handlers of other machines). But I suspect that this CheckOnly is never set here.

AD1024: Let's keep it as is for now. I may change it later! Thanks!

PR #861, opened 2025-05-14
Add Java equality and hash functions to machines in pobser…

lewisbru:
…ve mode

Generate deepEquals(), equals(), and hashCode() methods in the Java code for machines when compiled in pobserve mode.

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/Java/EventGenerator.cs
+++ b/Src/PCompiler/CompilerCore/Backend/Java/EventGenerator.cs
@@ -78,6 +78,45 @@ namespace Plang.Compiler.Backend.Java
                 WriteLine($" return \"{eventName}\"; }}");
             }
 
+            WriteLine();
+            Write("public int hashCode() {");
+            if (hasPayload)
+            {
+                WriteLine(" return payload.hashCode(); }");
+            }
+            else
+            {
+                WriteLine(" return 0; }");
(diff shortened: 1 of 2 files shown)


ankushdesai (commented):
I dont understand why we need these? 

Event equality can be done at the base class level itself as it has access to the event name and payload?

I dont think we use machine equality with fields anywhere?

lewisbru:
This is to enable the structural comparison we wanted to add for checker state validation in PObserve.

Src/PCompiler/CompilerCore/Backend/Java/EventGenerator.cs
```diff
@@ -78,6 +78,45 @@ private void WriteEventDecl(Event e)
...
+            }
+            else
+            {
+                WriteLine("return true; }");
+            }
+
+            WriteLine();
+            WriteLine("public boolean equals(Object other) {");
+            Write("return this.getClass() == other.getClass()");
+            if (hasPayload)
```
ankushdesai, line 113: May be I am missing something here, why can we not achieve this at the base class level?

lewisbru: My mistake, it looks like it is handled in the base class.

ankushdesai (approved):

PR #865, opened 2025-05-21
[fix] Postpone type checking of invariants, axioms, init-conditions and pure functions

AD1024:
This PR fixes the issue where the compiler raises type checking errors when the projection of a machine/event is used before their declarations are populated.

@FedericoAureliano Please take a look and let me know if there are other types of declarations need similar changes. Thanks!


ankushdesai (approved):

AD1024:
Additional fixes: proof blocks are handled last since invariant blocks may not be populated.

ankushdesai (approved):

PR #879, opened 2025-06-30
Dev p3.0/param testcases

ChristineZh0u:
Merging param test cases branch into major 3.0


.github/workflows/macosci.yml
```diff
@@ -24,4 +24,10 @@ jobs:
     - name: Build
       run: dotnet build --configuration Release
     - name: Test
-      run: dotnet test --configuration Release
+      run: dotnet test --configuration Release --blame-crash --blame-hang --blame-hang-timeout 300s --logger:"console;verbosity=detailed"
```
ankushdesai, line 27: Can you please explain this change?

Src/PChecker/CheckerCore/SystematicTesting/TestingEngine.cs
```diff
@@ -190,7 +190,7 @@ public static TestingEngine Create(CheckerConfiguration checkerConfiguration, As
                         Console.Out.WriteLine($"{mi.DeclaringType.Name}");
                     }
 
-                    Environment.Exit(0);
```
ankushdesai, line 193: Why this change?

Src/PCompiler/CompilerCore/Backend/PChecker/PCheckerCodeGenerator.cs
```diff
@@ -21,7 +22,8 @@ public class PCheckerCodeGenerator : ICodeGenerator
         /// This compiler has a compilation stage.
         /// </summary>
         public bool HasCompilationStage => true;
-        private static List<Variable> _globalParams = [];
+        [ThreadStatic]
+        private static List<Variable> _globalParams;
```
ankushdesai, line 26: why this change?

Src/PCompiler/CompilerCore/Backend/PChecker/PCheckerCodeGenerator.cs
```diff
@@ -102,6 +104,7 @@ private CompiledFile GenerateSource(CompilationContext context, Scope globalScop
             // write the top level declarations
             foreach (var decl in globalScope.AllDecls)
             {
+                // Console.WriteLine("top-level decl: " + decl.Name);
```
ankushdesai, line 107: Please make a pass to remove unwanted changes.

ankushdesai (changes requested):

.github/workflows/macosci.yml
ChristineZh0u: This change is to get better logs when the GitHub Action fails. Without it, we just see that something crashed, but not what caused it. With these flags, we get more detailed output and can tell which test hung or crashed, which makes debugging a lot easier.

Src/PChecker/CheckerCore/Checker.cs
```diff
@@ -51,7 +51,12 @@ public static void Run(CheckerConfiguration configuration)
             switch (configuration.Mode)
             {
                 case CheckerMode.BugFinding:
-                    TestingProcess.Create(configuration).Run();
+                    var testingEngine = TestingProcess.Create(configuration);
```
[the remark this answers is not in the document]
ChristineZh0u: We still call TestingProcess.Create() even when configuration.ListTestCases is true because the logic to list test cases lives inside the Create() method. That method is responsible for discovering and printing the available test cases, so we need to call it regardless.

The reason for making this change now is that the original implementation used Environment.Exit to terminate after listing test cases. This caused issues when running tests in-process because Environment.Exit would terminate the entire test run, not just the individual test. To fix this, I replaced the exit call with a simple return, which exits the method cleanly without affecting other tests.

Src/PCompiler/CompilerCore/Backend/PChecker/PCheckerCodeGenerator.cs
ChristineZh0u: This is a global variable which can't be made non-static, otherwise it defeats the purpose of global variable.

Tst/RegressionTests/Feature1SMLevelDecls/Correct/ParamTest/ParamTest.p
```diff
@@ -0,0 +1,33 @@
+param g1 : int;
```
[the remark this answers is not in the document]
ChristineZh0u: added a new param (g5) for boolean

lewisbru (approved):

PR #885, opened 2025-07-22
P Verifier

FedericoAureliano:
Add new verification backend. See `Docs/docs/advanced/PVerifierLanguageExtensions` for documentation, an announcement, a tutorial example, and install instructions.


Src/PCompiler/CompilerCore/CompilerConfiguration.cs
```diff
@@ -154,6 +158,10 @@ public CompilerConfiguration(ICompilerOutput output, DirectoryInfo outputDir, IL
         /// Gets or sets a value indicating whether debug information should be included in output.
         /// </summary>
         public bool Debug { get; set; }
+        public int Timeout { get; set; }
```
ankushdesai, line 161: Can we pull these configuration into a separate class and make then `PVerifierConfig`

Src/PCompiler/CompilerCore/TypeChecker/Analyzer.cs
```diff
@@ -32,7 +34,67 @@ public static Scope AnalyzeCompilationUnit(ICompilerConfiguration config,
                 FunctionValidator.CheckAllPathsReturn(handler, machineFunction);
             }
 
-            // Step 2: Validate no static handlers
+            // Step 3b: for PVerifier, fill in body of Invariants, Axioms, Init conditions and Pure functions and functions with pre/post conditions
```
ankushdesai, line 37: Can we make sure this is done only for the PVerifier mode?

ankushdesai (changes requested):

PR #934, opened 2026-02-13
[PVerifier] Improve proof speed by not generating unused procedures

TheKK:
I found that there are  unused procedures stand inside `.ucl` files generated by PVerifier backend and we get visible speed up (about 10% for Advanced/4_Paxos) in time after removing them.

This speed up should be positively correlated to the number of handlers of the entire system, thus give us better experience in the iteration of proofing.

I'll add some number later and also want to make sure that I didn't miss anything about `call` dependencies for entry/event_handler in `.ucl` files.

Please let me know if there's anything I could do better :)


TheKK:
```
1_ChainReplicationVerification
Benchmark 1: mod-p compile
  Time (mean ± σ):     28.241 s ±  1.203 s    [User: 177.758 s, System: 9.095 s]
  Range (min … max):   26.951 s … 30.051 s    5 runs

Benchmark 2: p compile
  Time (mean ± σ):     31.990 s ±  1.039 s    [User: 188.420 s, System: 9.229 s]
  Range (min … max):   30.760 s … 32.872 s    5 runs

Summary
  mod-p compile ran
    1.13 ± 0.06 times faster than p compile
```

ankushdesai:
Nice find — the speedups are very real and consistent. Two things to consider before this is ready:

**1. Federico's dedup question.** In the current diff this is actually not triggerable: `GenerateEntryHandler` only ever appends `s.Entry` (the one `else` branch), and `GenerateEventHandler` returns at most one element from its switch. So per `GenerateMain` call the list has 0 or 1 items. That said, switching `List<Function>` → `HashSet<Function>` is cheap insurance and makes the dedup intent explicit — especially relevant for point 2.

**2. Transitive callees — possible correctness gap.** `GenerateMachineProcedures` previously emitted *every* entry in `m.Methods`, which includes both anonymous handler bodies and named machine-scoped `fun`s. The new code only emits the handler's direct callee (`s.Entry` or `EventDoAction.Target`). But that callee's body can itself contain `FunCallStmt`s, which the codegen lowers unconditionally at `Uclid5CodeGenerator.cs:1801-1804` as `call <other_fun>(...)`. If the entry function calls a named helper, the helper's procedure is no longer emitted and Uclid will hit an unresolved call.

Sketch:
\`\`\`p
machine M {
  fun helper() { /* ... */ }
  state Init { entry { helper(); } }
}
\`\`\`
Before this PR, \`helper\` was always emitted (it's in \`m.Methods\`). After, only \`s.Entry\` is — so \`call helper(...)\` inside the entry body is dangling.

Why the benchmarks still all pass: I checked \`Tutorial/Advanced/*Verification\` and none of them declare machine-scoped named \`fun\`s — the only \`fun\` declarations are top-level globals (\`RandomInt\`, \`RandomParticipant\`) which flow through \`GenerateGlobalProcedures\`, a separate code path. So this benchmark set never exercises machine-method → machine-method calls; the perf wins are sound but the test surface doesn't catch the issue.

Suggested fix: compute the transitive call closure from the handler's direct callee — walk each function's body collecting \`FunCallStmt\` targets whose \`Function.Owner == machine\`, until fixpoint. A \`HashSet<Function>\` naturally handles both this and Federico's dedup point. Adding a small regression test with a machine helper that calls another helper would lock the invariant in.

Minor nit while you're in there: \`functionDependeicies\` typo in \`GenerateMain\` (spelled correctly as \`functionDependencies\` inside \`GenerateEntryHandler\`).

PR #941, opened 2026-03-26
Add @Generated annotation to PObserve generated Java classes

ylecaillez:
**NOTE:** we do have a post compile script which add these @Generated annotation at build time but maybe you'll be interested to have this embedded directly in the compiler.

Emit @Generated(value="P Compiler", date=...) from javax.annotation.processing.Generated on the top-level PTypes, PEvents, and PMachines classes so that tools and IDEs can identify generated code.

Also unify the three separate DateTime.Now calls (DoNotEditWarning, FFI banner, and the new annotation) into a single GenerationTimestamp captured once at class-load time, ensuring all output files carry the same timestamp.

**Example**

```
/***************************************************************************
 * This file was auto-generated on Saturday, 11 April 2026 at 05:20:27 UTC.
 * Please do not edit manually!
 **************************************************************************/
...
@Generated(value = "P Compiler", date = "2026-04-11T05:20:27Z")
public class PMachines {
 ```
 
  ```
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
% 
% P <-> Java Foreign Function Interface Stubs
% 
% This file was auto-generated on Saturday, 11 April 2026 at 05:20:27 UTC.
%
```

### DIFF
--- a/Src/PCompiler/CompilerCore/Backend/PObserve/Constants.cs
+++ b/Src/PCompiler/CompilerCore/Backend/PObserve/Constants.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Text;
 
@@ -13,6 +14,14 @@ namespace Plang.Compiler.Backend.Java
     /// </summary>
     internal static class Constants
     {
+        /// <summary>
+        /// A single UTC timestamp captured at class load time, shared by all generated file headers
+        /// and annotations to ensure consistency across all output files.
+        /// </summary>
+        private static readonly DateTime GenerationTimestampUtc = DateTime.UtcNow;
+
+        private static readonly CultureInfo Inv = CultureInfo.InvariantCulture;
+
         #region P Java runtime constants
 
         public static readonly string PRTNamespaceName = "pobserve.runtime";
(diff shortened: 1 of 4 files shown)


ankushdesai:
Thanks for the PR! One suggestion: I'd vote to drop the timestamp entirely — both the `date=` field on `@Generated` and the existing wall-clock timestamps in the FFI banner and `DoNotEditWarning` banner.

Reasoning: embedding `DateTime.UtcNow` in generated source means `p compile` produces different bytes on every run, even when no input changed. That breaks content-hash-based caching (Bazel, container layer caches, etc.), inflates incremental rebuild scope, and makes diffs of regenerated output noisy without adding information a reader can actually use (the git commit and the compiler version tell you "when" more reliably). JSR-269 makes `date` optional precisely for this reason — `@Generated(value = "P Compiler")` on its own is the common shape and gives IDEs and lints everything they need.

If keeping the banners' human-readable "this is generated" message is desirable, the wording works fine without a timestamp: just `"This file was auto-generated by the P compiler. Please do not edit manually."`

Concretely this would simplify the diff a lot — the whole `GenerationTimestampUtc` / `Inv` plumbing in `Constants.cs` goes away, and the four `WriteLine(Constants.GeneratedAnnotation)` call sites just emit `@Generated(value = "P Compiler")`.

Bonus: if you'd ever want a timestamp back later, the reproducible-builds convention is to honor a `SOURCE_DATE_EPOCH` env var when set, falling back to no timestamp otherwise — that's an easy opt-in for users who genuinely want it.

(Separately, the Copilot review comment about UTC vs local-time mismatch looks stale — the current diff uses UTC consistently for both the annotation and the banners. Worth dismissing.)

PR #969, opened 2026-05-28
PeasyAI: default to collecting-mode compilation in subprocess.run

ankushdesai:
## Summary

PeasyAI's compile path shells out to `p compile` via three `subprocess.run` sites — none of which previously propagated the `P_COMPILER_COLLECT_ERRORS` env var. With P compiler 3.0+'s new multi-error mode (#957/#963/#965), enabling the env var lets PeasyAI's `peasy-ai-fix-all` loop converge in **O(N/k) LLM round trips instead of O(N)** where k is the average errors-per-iteration. This is the motivating use case for the entire multi-error initiative.

## Change

At each of three subprocess.run sites:

```python
env = os.environ.copy()
env.setdefault("P_COMPILER_COLLECT_ERRORS", "1")
result = subprocess.run([...], ..., env=env)
```

`setdefault` (not assignment) preserves the user's explicit override:
- **Unset** → child sees `"1"` (collecting mode, new default)
- **Set to `"0"` / `"false"`** → child sees the user's value (strict mode)
- **Set to anything else** → child sees the user's value

## Sites changed

- `Src/PeasyAI/src/core/services/compilation.py` — canonical compile entry point used by `peasy-ai-compile` MCP tool + `fix_iteratively` loop
- `Src/PeasyAI/src/utils/compile_utils.py` — legacy helper for evaluation pipelines
- `Src/PeasyAI/src/core/services/generation.py:1445` — candidate-scoring compile during code generation

## Not changed

- `Src/PeasyAI/src/utils/checker_utils.py` — invokes `p check`, not `p compile`; env var has no effect there
- `Src/PeasyAI/src/core/compilation/environment.py` — toolchain validation calls (`dotnet --list-sdks` etc.), not P compilation
- The error parser already supports multiple errors via `compilation.get_all_errors` and `error_parser.parse` (both iterate all matches with `re.finditer`). No parser changes needed.

## Follow-up (separate PR)

The multi-agent audit recommended updating `fix_iteratively` to call `get_all_errors` instead of `parse_error` and send all errors to the LLM in one batch. That's a behavioral change to the fix loop. This PR is the prerequisite (env var must propagate first); the batching work is a separate PR.

## Test plan

- [ ] CI green (Python tests under `Src/PeasyAI/tests/`)
- [ ] Manual: run `peasy-ai-fix-all` on a P file with 3 independent errors; observe that the LLM sees all 3 in one round trip

### DIFF
--- a/Src/PeasyAI/src/core/services/compilation.py
+++ b/Src/PeasyAI/src/core/services/compilation.py
@@ -5,6 +5,7 @@ Handles P project compilation and PChecker execution.
 This service is UI-agnostic.
 """
 
+import os
 import subprocess
 import logging
 import traceback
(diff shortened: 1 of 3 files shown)


ankushdesai:
Superseded by #970, which makes multi-error mode the compiler default. PeasyAI no longer needs to propagate the env var because the compiler defaults to collecting mode for all callers. `fix-all` now converges in O(N/k) iterations automatically without any PeasyAI-side changes.

PR #974, opened 2026-06-18
Agentic invariant-mining toolkit for P (propose → check → repair → judge)

ankushdesai:
## What

Standalone tooling that uses an LLM to **propose** intent-level specification invariants for P programs, has **PChecker verify every one** (the sound oracle), and then **judges** each result. No dependency on the PInfer C# backend — these are pure utility scripts driving `p compile` / `p check`.

The loop:

```
agent proposes (from source + design intent)
   → prep_candidates.py   (deterministic P fixups)
   → repair loop          (LLM fixes non-compiling specs from the exact compiler error)
   → check_candidates.py  (model-check each: HOLDS / FAILS+cex / VACUOUS / INCONCLUSIVE)
   → agent judges          (verified / vacuous / bug / spurious, grounded in the cex)
```

## Files (`Src/PInfer/Scripts/`)

| File | Role |
|---|---|
| `check_candidates.py` | wire candidate monitors into a project, generate one test each, model-check, classify; vacuity canaries; flags pre-existing SUT failures as `INCONCLUSIVE` (not a false bug) |
| `prep_candidates.py` | deterministic fixups for LLM-generated P (entity unescape, `var`-decl hoisting, `!in`/`for` rewrites) |
| `diag_compile.py` | compile each candidate alone, capture its first compiler error |
| `repair_workflow.js` | LLM-repair Workflow: rewrite a non-compiling spec given the exact error + P syntax rules (iterable) |
| `verify_repairs.py` | recompile repaired specs to measure recovery |
| `README_agentic_invariants.md` | how the pipeline fits together |

## Why this shape

- **Propose from intent, not the implementation** — mirroring the code makes a bug look like a correct invariant. The gap between intent and implementation is where bugs live.
- **No trace pre-filter** — a failing candidate is *information*: either a required property the system violates (bug) or an over-strong candidate. Only a sound check + a judgment over the concrete counterexample tells them apart.
- **Vacuity is checked** — a `<Name>_canary` companion reveals a monitor whose assertion never fires.

## Validation (4 tutorial benchmarks)

- 82 proposed candidates → **50 machine-checked verified invariants** (including rediscoveries of the authors' own specs plus many they didn't write).
- The judge layer adjudicated 15 model-checker failures as **0 confirmed bugs, 6 monitor-encoding bugs, 9 spurious** — a naive "fails = bug" pipeline would have filed 15 false reports.
- The compile-and-repair loop recovered **all 23** previously-uncompilable candidates in two rounds (compile yield ~72% → ~100%).

## Scope / status

Experimental utility tooling, portable dotnet/`p` resolution (Linux + macOS). Opened for discussion alongside the PInfer effort; the propose/judge steps are LLM-driven (run with Claude Code or any agent), this PR is the deterministic middle.


ankushdesai:
Superseded by #977, which carries the hardened successors of these scripts (frozen candidate schema, auto-canary vacuity, static name gate, baseline-differential attribution, ranking) plus the design plan (Src/PInfer/Scripts/PLAN.md).

PR #982, opened 2026-07-16
Harden compiler literal parsing: located diagnostic instead of crash

ankushdesai:
## Summary

Integer and float **literals** (and tuple field indices) were parsed straight from the ANTLR lexeme with `int.Parse` / `double.Parse` and **no guard**. This had two silent-failure consequences:

- An out-of-range literal (`a = 9999999999999;`) or an oversized tuple index (`t.99999999999`) threw `OverflowException`.
- On a comma-decimal host locale (de-DE, fr-FR, …), `double.Parse` misparsed or rejected a valid P float.

Those are **not** `TranslationException`s, so they escaped `Analyzer.TolerantStep`'s collecting-diagnostics machinery and **aborted the entire compilation with an opaque .NET stack trace** — instead of the clean, located error the P 3.0 multi-error architecture promises, and taking every other diagnostic down with them.

## Change

- New `LiteralParsingUtils` with culture-invariant `TryParseIntLiteral` / `TryParseFloatLiteral` that return `false` instead of throwing.
- All **10** parse sites routed through it, each using the idiom its file already follows:
  - **`ExprVisitor`** (int literal, decimal float, two tuple-access indices) → collecting idiom: `handler.Diagnostics.Report(handler.ValueOutOfRange(...))` + `return new ErrorExpr(context)`.
  - **`ParamVisitor`** (5 sites), **`DeclarationVisitor`** (enum value), **`ModuleExprVisitor`** (t-wise count) → throw idiom: `throw handler.ValueOutOfRange(...)` (a `TranslationException`).
- Invariant culture also fixes the float **locale** bug at both float sites.

## Tests

Two `StaticError` fixtures (the validator fails on a thrown exception and passes only on exit code 1 — so these directly guard "clean diagnostic, not crash"):

- `Tst/RegressionTests/Feature3Exprs/StaticError/IntLiteralOutOfRange`
- `Tst/RegressionTests/Feature3Exprs/StaticError/TupleIndexOutOfRange`

Both now emit `value is out of range for type 'int'` with a source location.

## Verification

Full Release regression suite green — `dotnet build/test --configuration Release`, **726/726 passed, 0 failures** (same steps as `ubuntuci.yml`).

### DIFF
--- a/Src/PCompiler/CompilerCore/TypeChecker/DeclarationVisitor.cs
+++ b/Src/PCompiler/CompilerCore/TypeChecker/DeclarationVisitor.cs
@@ -248,7 +248,11 @@ namespace Plang.Compiler.TypeChecker
         {
             // name=iden ASSIGN value=IntLiteral
             var elem = (EnumElem) nodesToDeclarations.Get(context);
-            elem.Value = int.Parse(context.value.Text);
+            if (!LiteralParsingUtils.TryParseIntLiteral(context.value.Text, out var enumValue))
+            {
+                throw Handler.ValueOutOfRange(context, "int");
+            }
+            elem.Value = enumValue;
             return elem;
         }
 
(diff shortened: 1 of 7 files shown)


ankushdesai:
Thanks for the review — both points addressed in `73754fdd6`:
1. `TryParseIntLiteral` now uses `NumberStyles.None` (digits only) to match the grammar's `IntLiteral : [0-9]+`.
2. The tuple-lvalue out-of-range diagnostic is now reported at `context.@int()` for a precise source span.

Both regression fixtures still produce a located `value is out of range for type 'int'` diagnostic (exit code 1).

PR #983, opened 2026-07-16
Add official Docker image for the P toolchain

### LINKED ISSUE #980: Request: official P docker image
P requires several toolchains to compile and check models — the .NET SDK (P compiler + PChecker) and Java + Maven (PSym/PExplicit backends). Setting these up is a barrier for getting started and for CI.

Proposal: Publish an official Docker image bundling the .NET SDK, JDK + Maven, graphviz, and the p CLI pre-installed, built and pushed via a GitHub release action.

Starting point:
```Dockerfile
FROM mcr.microsoft.com/dotnet/sdk:8.0

# Java (for PSym / PExplicit backends) + Maven + graphviz (visualization)
RUN apt-get update \
    && apt-get install -y --no-install-recommends \
        openjdk-17-jdk-headless \
        maven \
        graphviz \
    && rm -rf /var/lib/apt/lists/*

ENV JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64

# Install the P CLI as a .NET global tool (pin the version)
ARG P_VERSION=2.2.6
RUN dotnet tool install --global P --version ${P_VERSION}
ENV PATH="${PATH}:/root/.dotnet/tools"

# Sanity check
RUN p --help

WORKDIR /workspace
CMD ["bash"]
```

If this looks interesting to you all, I'm happy to put together a PR.

crowecawcaw:
Closes #980.

Adds an official Docker image for the P toolchain so P can be compiled and checked without installing the .NET SDK, Java, Maven, and graphviz by hand — especially handy for CI and for trying P out.

## What's included

- **`Dockerfile`** — multi-stage build.
  - Stage 1 builds the `p` CLI **from this repository's source** (`dotnet pack`), so the image always matches the committed/released code (no dependency on the NuGet package already being published).
  - Stage 2 is the toolchain image: **.NET SDK 8.0, JDK 17, Maven, graphviz**, and the `p` tool installed from the locally-built package.
  - JDK **17** is used deliberately: `Src/PEx/pom.xml` targets Java 17, so the PEx/PSym backends need a JVM that can run Java-17 bytecode.
  - `JAVA_HOME` is resolved via a stable symlink so it's correct on both amd64 and arm64 (the Debian JDK path is arch-specific).
- **`.dockerignore`** — keeps the build context small (excludes `.git`, build drops, docs, etc.).
- **`.github/workflows/docker-image.yml`** — PR CI. Path-filtered so it only runs when the `Dockerfile`, `.dockerignore`, or the workflow itself changes (plus manual dispatch) rather than on every push. Builds **natively** on amd64 (`ubuntu-latest`) and arm64 (`ubuntu-24.04-arm`), smoke-tests `p compile` + `p check` on a tutorial inside the container, and runs a Trivy vulnerability scan (report-only).
- **`.github/workflows/docker-publish.yml`** — on a published GitHub Release, builds per-arch natively, pushes each by digest, and assembles a **multi-arch manifest** to `ghcr.io/p-org/p` tagged with the release version + `latest`.
- **`.github/dependabot.yml`** — keeps the base image and the GitHub Actions used here up to date.
- **Docs** — a Docker section in the [installation guide](Docs/docs/getstarted/install.md) (referenced from the quick start), plus a README badge and run snippet.

## Usage

```shell
docker run --rm -it -v "$PWD":/workspace ghcr.io/p-org/p:latest
# inside the container:
p compile
p check -tc <testcase> -i 1000
```

## Testing done

Built and ran locally on arm64:
- `docker build` succeeds.
- Inside the container: `p compile` + `p check -tc tcSingleClient -i 200` runs the Java PEx backend end-to-end (200 schedules explored, 0 bugs, coverage report written); JDK 17, Maven, graphviz, and `JAVA_HOME` all verified in both login and non-login shells.

CI in this PR exercises the build + smoke test on both amd64 and arm64.

## Note on availability

The image is only published when a GitHub Release is created, and a repo's first GHCR package is created **private**. So before anyone can `docker pull ghcr.io/p-org/p`, a maintainer will need to:

1. Cut a release (or run the publish workflow manually) to push the first image, and
2. Flip the `ghcr.io/p-org/p` package to **public** (optionally linking it to the repo).

No code change needed for that — it's a one-time settings step around the first release.

## Choices made (happy to change)

- **Version tag source** — the publish workflow derives the image version from the release's git tag (`p-<version>`), plus `latest`. Releases cut without that tag format would only get `latest`.
- **Trivy is report-only** (`exit-code: 0`) so new upstream CVEs don't block PRs. Can be made blocking (keeping `ignore-unfixed: true`) if preferred.
- **Image scope** — this is a toolchain/dev image (SDK-based, not slim). A smaller runtime image could be a separate tagged variant later.
- Kept the image and workflows close to the Dockerfile proposed in #980.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,0 +1,125 @@
+name: Publish Docker image
+
+# Build the multi-arch P toolchain image and push it to the GitHub Container
+# Registry (GHCR) when a GitHub Release is published. Can also be run manually
+# to (re)publish for a given tag.
+on:
+  release:
+    types: [ published ]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Image tag to publish (e.g. 2.2.6). Defaults to 'latest' only."
+        required: false
+        default: ""
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}   # -> ghcr.io/p-org/p
+
+jobs:
+  # Build each architecture natively (no QEMU) and push it by digest.
+  build:
+    name: Build & push (${{ matrix.arch }})
+    runs-on: ${{ matrix.runner }}
+    permissions:
+      contents: read
+      packages: write
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: amd64
+            runner: ubuntu-latest
+          - arch: arm64
+            runner: ubuntu-24.04-arm
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+
+      - name: Log in to GHCR
+        uses: docker/login-action@v4
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push by digest
+        id: build
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          platforms: linux/${{ matrix.arch }}
+          # Push an untagged, by-digest image; the manifest list is assembled
+          # in the merge job below.
+          outputs: type=image,name=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }},push-by-digest=true,name-canonical=true,push=true
+          cache-from: type=gha,scope=${{ matrix.arch }}
+          cache-to: type=gha,mode=max,scope=${{ matrix.arch }}
+
+      - name: Export the digest
+        run: |
+          mkdir -p /tmp/digests
+          digest="${{ steps.build.outputs.digest }}"
+          touch "/tmp/digests/${digest#sha256:}"
+
+      - name: Upload digest
+        uses: actions/upload-artifact@v4
+        with:
+          name: digest-${{ matrix.arch }}
+          path: /tmp/digests/*
+          if-no-files-found: error
+          retention-days: 1
+
+  # Assemble the per-arch digests into a single multi-arch manifest and tag it.
+  merge:
+    name: Create multi-arch manifest
+    runs-on: ubuntu-latest
+    needs: [ build ]
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Download digests
+        uses: actions/download-artifact@v4
+        with:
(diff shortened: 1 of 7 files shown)


crowecawcaw:
@ankushdesai wondering if you have time to take a look

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,125 @@
...
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Image tag to publish (e.g. 2.2.6). Defaults to 'latest' only."
+        required: false
+        default: ""
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}   # -> ghcr.io/p-org/p
```
AD1024: Should the repo name be a lower case for the docker image name? (since the repo name is "P", the image name may also contain that upper case P, which may cause an error).

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,125 @@
...
+      - name: Compute image tags
+        id: meta
+        uses: docker/metadata-action@v6
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          # On release: tag with the semver from the release tag (git tags are
+          # of the form `p-<version>`) plus a rolling `latest`. On manual runs:
+          # honor the optional `tag` input.
+          tags: |
+            type=match,pattern=p-(.*),group=1,event=tag
```
AD1024: I'm not quite familiar with this, but claude said `type=match` does not support the qualifier `event=tag`. Maybe it can be removed?

AD1024:
Hi @crowecawcaw 
Thanks for contributing. This is very useful for new users. I just approved the workflow and will see how it runs. I'm not quite familiar with github action configs, so my comments might not be accurate.

.github/workflows/docker-publish.yml
crowecawcaw: Yes, cleaned up these tags

crowecawcaw:
Made some minor fixes based on your comments. Also I triggered an image release on my fork with the changes which succeeded: https://github.com/crowecawcaw/P/actions/runs/30723809943

AD1024 (approved):
LGTM! Thanks for contributing the feature. Will merge after the CI run.

cc: @ankushdesai

PR #984, opened 2026-07-17
Fest: reproducibility & correctness fixes (timeline identity, RNG seeding, dead-code)

ankushdesai:
## Summary

Reproducibility & correctness fixes for the Fest feedback-guided testing subsystem, surfaced by a paper-vs-implementation review (`Fest-PChecker-review.md`). Three small, independent, verified changes (one commit each). No behavior change to the default (non-feedback) strategies.

### 1. Collision-free timeline identity (`5ede3dd05`)
Timeline identity was a **32-bit `String.GetHashCode`** of the canonical `<M,e1,e2>` abstract-timeline string, stored in `TestReport.ExploredTimelines` (`HashSet<int>`) and `FeedbackGuidedStrategy._visitedTimelines` (`HashSet<int>`). Hash collisions had two harmful effects:
- reported timeline coverage **undercounted** distinct timelines, and
- `ComputeDiversity`'s `_visitedTimelines` gate **wrongly discarded** a genuinely novel timeline whenever its 32-bit hash collided with a prior one.

Fix: key identity on the full canonical `GetAbstractTimeline()` string. `ExploredTimelines`/`_visitedTimelines` → `HashSet<string>` (matches the sibling `InternalErrors : HashSet<string>`; DataContract-serializes fine); drop the now-unused `GetTimelineHash()`.

### 2. Seed the feedback explore/exploit RNG (`23e16a04d`)
`FeedbackGuidedStrategy._rnd` was `new System.Random()` (clock-seeded), so the explore/exploit coin flip in `PrepareNextInput` diverged run-to-run even under a fixed `--seed` — a hole in the reproducibility guarantee for the feedback-guided strategies. Seed it from `CheckerConfiguration.RandomGeneratorSeed` in the constructor, mirroring `ControlledRandom` (still random when no seed is given).

### 3. Remove dangling `feedbackpctcp` + dead `BehavioralObserver` (`d78f5aeeb`)
- The seed-print list named a `feedbackpctcp` strategy with no construction branch (and not in the CLI sanitize allow-list), so selecting it would leave `Strategy` null. Removed the dead reference.
- `BehavioralObserver` was a fully-static coverage observer fed on every send/dequeue and per iteration, whose score was never wired into feedback/priority; it also used an unseeded MinHash and unbounded static state. Removed the class + its three call sites. `VectorTime` is intentionally kept (it's carried in `EventInfo`/`EventQueue` and merged on dequeue — not dead); pruning the now-unconsumed vector-clock maintenance is left to a follow-up.

## Verification
Full regression suite green after each step: **`dotnet test -c Release` → 722/722, 0 failures**.


ankushdesai:
Good catch — addressed in `17693a57d`. `GetTimelineMinhash` now hashes each timeline tuple with a stable FNV-1a hash (over UTF-8 bytes, field-separated) instead of `GetHashCode()`, so the minhash — and therefore `ComputeDiversity`/generator prioritization — is reproducible under a fixed `--seed` across processes. This complements the `_rnd` seeding in the same PR.

